PR #86, opened 2014-08-26
Allow upstream apt source pin priority to be changed, or eliminated

RsrchBoy:
Apt pins are...  Interesting.  The hardcoded "10" was sufficient to keep my
system from allowing me to update lxc-docker via the expected means.

This changeset allows one to override the default pin priority level via
`apt_source_pin_level`, or even disable it completely via

``` pin_upstream_package_source```.
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -68,6 +68,14 @@
 #   Won't install or define the docker package, useful if you want to use your own package
 #   Defaults to true
 #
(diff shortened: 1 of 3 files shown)


garethr:
I've merged https://github.com/garethr/garethr-docker/pull/570 which addresses this, so closing this PR. Thanks for raising the issue.

Issue #91, opened 2014-09-09
Add service_provider parameter to docker class.


mtbvang:
I'm hitting the issues mentioned in the following post and it seems that the only work around at the moment is to add a service_provider parameter to the docker class to enable it to be settable. I'm happy to make the changes and do pull request if you think that is a good idea.

 http://ask.puppetlabs.com/question/13635/how-to-use-puppet-modules-that-require-upstart-with-docker/

Docker containers don't really support the puppet upstart service provider and the only work around I know of at the moment is to set provider=base. Seems like a good idea to expose the provider as a parameter anyway.

garethr:
Hi @mtbvang. First up, apologies for not reply to this much sooner. I'm just going through triaging all the issues. The module does now support passing a provider directly - https://github.com/garethr/garethr-docker/blob/master/manifests/init.pp#L300. Thanks.

PR #95, opened 2014-09-13
Added service_provider parameter to docker and service classes.

mtbvang:
Fix for issue #91 that was raised to add a service_provider parameter to docker and service classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ pkg
 spec/fixtures
 .bundle/
 .vagrant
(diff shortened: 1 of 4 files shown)


garethr:
Thanks @mtbvang. I think this is now cleaned up in the current master. In particular with https://github.com/garethr/garethr-docker/commit/9f615c6a1e03de4a8ab274d003e82f188663df54. If that's not the case please do reopen this issue.

PR #171, opened 2014-12-22
patch to garethr-docker#165

tfhartmann:
Ignoring my .ruby-versions and vim swap files

Adding tests

We don't need a test for the run define

Adding ability to pass shell values into init cfg files

This should allow you to pass an array of shell commands to the
end of the sysconfig/docker files


garethr:
@tfhartmann thanks for adding this feature and the tests

PR #172, opened 2014-12-30
docker::run - add Restart support to systemd unit file

bwalex:
I've also updated the upstart file, but as far as I can see, that one isn't actually used anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ Specifying `pull_on_start` will pull the image before each time it is started.
 
 The `depends` option allows expressing containers that must be started before. This affects the generation of the init.d/systemd script.
 
+The service file created for systemd and upstart based systems enables automatic restarting of the service on failure by default.
+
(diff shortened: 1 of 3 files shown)


garethr:
@bwalex is there a reason for this to not be the default? I'm thinking that we wouldn't need the logic and the new flag and we could just add the lines into the systemd config file. This feels like it would be useful in all cases?

bwalex:
It's certainly useful to me, and I can't think of a particularly good reason for it not to be default, other than the usual "someone might not want it". If you think it should be default, and there shouldn't even be a flag to disable it, I can rework the PR to do just that.

garethr:
@bwalex I think that's the right way to go to start with. We can always make it configurable later depending on usage, but it feels like a good sane starting point in any case. Thanks

bwalex:
That's that then, I think.

garethr:
@bwalex thanks, this will go out in next release which should be shortly

PR #173, opened 2014-12-31
bring sysconfig template for RHEL7 and Fedora up-to-date for DNS

nrvale0:
The template for RHEL7/CentOS7/Fedora was not setting DNS servers and search path in /etc/sysconfig/docker. This template is very similar to the one used on other platforms and the two could/should probably be consolidated. 

Sorry, but I'm in a crunch so I just "modernized" the CentOS7/etc version as I didn't have time to dive into the spec and acceptance testing and make the proper prep for merging the two templates and testing the conditional logic. FWIW, I have tested this branch on a live VM I've been using for testing of Jenkins + Docker work and things look fine.

### DIFF
--- a/templates/etc/sysconfig/docker.rhel7.erb
+++ b/templates/etc/sysconfig/docker.rhel7.erb
@@ -1,4 +1,19 @@
 # This file is managed by Puppet and local changes
 # may be overwritten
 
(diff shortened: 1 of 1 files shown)


garethr:
@nrvale0 no problem, and thanks for addressing. Will look at killing some of the duplication in the future.

PR #174, opened 2015-01-02
Reusable function to render flags for `docker run`

mcls:
This will make it easier to add/change flags


garethr:
@mcls excellent, thanks, much appreciated.

PR #179, opened 2015-01-07
Define docker::exec{} for garethr-docker#178

aries1980:


garethr:
Thanks, this looks great. Thanks for the addition.

PR #183, opened 2015-01-15
[Feature] Load images from tar file

paschdan:
sometimes one want to load images that were exported via 
```
docker save -o myimage.tar myimage
```
to get loaded with 

```
docker load -i myimage.tar
```

This PR achieves this


garethr:
@paschdan thanks, this is great. Much appreciated.

PR #187, opened 2015-02-06
Export HOME environment variable in init script.

lmovsesjan:
HOME=/root in current init script doesn't work and it requires to be called as export HOME=/root.


garethr:
Hi @lmovsesjan, it looks like this PR has pulled in lots of other commits. It currently doesn't merge cleanly and it's unclear (due to all the commits and merges with other branches) what the change is. If you could rebase this against master that would be useful. Thanks

PR #188, opened 2015-02-12
add containment of subclasses into main docker class

brunoleon:
I encounterd an issue with the following code which ensures the disk is properly formatted and mounted before installing docker:
  Profile::Disk::App[$app_name] ->
  class { '::docker': }
Actually Puppet was still installing docker before the disk is ready (docker::install class was applied too early)

This is solved by using containment feature.


garethr:
Hi @brunoleon. Thanks for this fix, much appreciated.

PR #190, opened 2015-02-26
Detect when image id changes and restart containers

shadone:
Automatically restart container service if the image id has changed.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -163,6 +163,14 @@ define docker::run(
 
   if $restart_service {
     File[$initscript] ~> Service["docker-${sanitised_title}"]
+
+    exec { "check-whether-image-id-changed-${sanitised_title}":
+      command => '/bin/true',
+      onlyif  => "test \"$(service docker-${sanitised_title} imageid)\" != \"$(docker inspect -f \'{{.Id}}\' $image)\"",
+      path    => ['/sbin', '/bin', '/usr/sbin', '/usr/bin'],
+      notify  => Service["docker-${sanitised_title}"],
+      require => File[$initscript],
+    }
   }
   else {
     File[$initscript] -> Service["docker-${sanitised_title}"]
--- a/templates/etc/init.d/docker-run.erb
+++ b/templates/etc/init.d/docker-run.erb
@@ -135,6 +135,12 @@ case "$1" in
     fi
     docker inspect $(cat $cidfile)
     ;;
+    imageid)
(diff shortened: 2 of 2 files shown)


shadone:
There are no tests and to be honest I have no idea how to implement one. If you are potentially interested in this pull request I'll try my best to test it.

jovandeginste:
Am I correct that this will only restart the container while you run puppet agent/apply (not 'if' but 'while')? Will the exec part run every time the puppet agent is parsing the code, or only if anything changes elsewhere?

I'm confused why you put the test in the exec, and not in the init-script; now if anything goed wrong in the imageid-part (eg. the docker-command is not 'docker'), the containers will be restarted every time this part is executed... Can you not make some kind of image-status target in init that exits 0 or 1 when the image is up-to-date or not, and use that in puppet? Even then we still need to be careful how to exit, and respond correctly in the puppet manifest - the containers should only restart when the image id is found on both sides of the equation AND found to be different...

manifests/run.pp
```diff
@@ -163,6 +163,14 @@
 
   if $restart_service {
     File[$initscript] ~> Service["docker-${sanitised_title}"]
+
+    exec { "check-whether-image-id-changed-${sanitised_title}":
+      command => '/bin/true',
+      onlyif  => "test \"$(service docker-${sanitised_title} imageid)\" != \"$(docker inspect -f \'{{.Id}}\' $image)\"",
```
garethr: docker maybe available elsewhere, you can see the use of `docker_command` elsewhere in the module.

templates/etc/init.d/docker-run.erb
```diff
@@ -135,6 +135,12 @@ case "$1" in
     fi
     docker inspect $(cat $cidfile)
     ;;
+    imageid)
+    if [ "true" != "$(docker inspect --format='{{.State.Running}}' $(cat $cidfile))"  ] ; then
```
garethr: Ditto above, in the init script context this can use the `$docker` variable.

garethr:
@shadone sorry for taking an age to review this suggestion. I've left a few quick comments. I also note that this only affects the init scripts and not the systemd scripts. Given the widespread use of systemd now I'd need to support that path too. The branch is also a little out-of-date now so if you could rebase with master that would be great.

As for testing, you could test this using the acceptance tests with beaker. I'm happy for that not to be the case, but mentioning for interest or if you really want to take a run at it.

Thanks again, and apologies for the delay.

Issue #192, opened 2015-03-06
Managing images and running containers using hiera?


ricoli:
Currently there is no puppet class for docker images or containers, so I wonder if there is any way to use hiera configuration to manage them? Afaik we can only use hiera with classes... Any thoughts?

garethr:
I think this is what you're looking for for containers: https://github.com/garethr/garethr-docker/blob/master/manifests/run_instance.pp

I'd be happy to see a PR for something similar (even better a bunch of examples) for images too.

ricoli:
Ah yes indeed, I missed that class... yeah could do the same for the images so they can be prefetched.

ricoli:
Leaving a temporary solution here for the time being - adding this to a manifest:
```
$dockerImages = hiera('docker::images', {})
create_resources('docker::image', $dockerImages)
```

garethr:
Fixed in #194

PR #194, opened 2015-03-11
added convenience class for images

ricoli:
As per https://github.com/garethr/garethr-docker/issues/192
I didn't add a spec for this as I'm facing gem versions problems, let me know if it's essential for this. I suspect it's not essential as I couldn't find a spec for run_instance.pp either, which is similar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,14 @@ docker::image { 'ubuntu':
 }
 ```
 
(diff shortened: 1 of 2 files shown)


ricoli:
ping @garethr

garethr:
@ricoli thanks for this fix, especially adding to the docs

PR #199, opened 2015-03-25
Added user group management

ajcrowe:
This added support for automatically adding users to the docker group to allow access to the daemon.

PR #200, opened 2015-03-25
Fixes Debian init script issue

ajcrowe:
This is a small fix to avoid the init script being deleted on Debian hosts which don't use Upstart.

This allows Debian 7 hosts with the backports kernel to use this module without issue provided the following parameters are set.

```
docker::package_name   => 'lxc-docker',
docker::service_name   => 'docker',
docker::docker_command => 'docker',
```

If you need any others changes let me know!

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -50,11 +50,18 @@ class docker::service (
       $hasstatus     = true
       $hasrestart    = false
 
-      file { '/etc/init.d/docker':
-          ensure => 'link',
-          target => '/lib/init/upstart-job',
-          force  => true,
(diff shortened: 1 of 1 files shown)


garethr:
Hi @ajcrowe. Thanks for this.

It looks like a commit from https://github.com/garethr/garethr-docker/pull/199/files made it's way into this PR, specifically the README change.

Once I merge this I'll try and add those params to the default params, and add Debian to the acceptance test suite. Thanks for your help.

ajcrowe:
Opps, I've removed that commit now.

PR #201, opened 2015-03-26
Add hostentries to pass --add-host flags to the docker container

gtmtech:
I need the ability to pass --add-host flags to the docker init script - this does that.


garethr:
@gtmtech thanks for adding this, much appreciated.

Issue #186, opened 2015-01-22
init script should still work when 'use_name' is defined


clkao:
Apparently it's still trying to determine the container id from the cid file:

    % service docker-registry status
    cat: /var/run/docker-registry.cid: No such file or directory
    Usage: docker inspect [OPTIONS] CONTAINER|IMAGE [CONTAINER|IMAGE...]

from:

```
docker::run { 'registry':
    image    => 'registry:0.9.1',
    use_name => true,
    depends  => ['rediscache'],
    links    => ['rediscache:rediscache'],
    ports    => ['127.0.0.1:5000:5000'],
}
```

diranged:
We ran into this bug as well... for now, we're using `use_name => False` to deal with it. :(

diranged:
@garethr, have you seen this behavior?

marji:
Well, `/var/run` is a symlink to `/run` and twhich is mounted as tmpfs (or get deleted) on boot.
Therefore, `/var/run/docker-registry.cid` does not exist after boot.

marji:
I was wrong in my previous comment. The cidfile gets explicitly deleted by the stop function, so it has nothing to do with /var/run not being persistent.

I created PR #203 which addresses this issue.

garethr:
While adding the ability to run things using docker's in-built process manager I ran into this issue, even with the changes in #203. Specifically stoping and restarting containers failed.

On closer inspection, we shouldn't be deleting the cid file at all. I've made some changes and simplified the init script in https://github.com/garethr/garethr-docker/commit/6f2c509cf39cac35afc11e09c4dd4c829119faab which should resolve this problem. I've also add containers using `use_name` to the acceptance tests.

Let me know if anyone runs into this again post the changes.

PR #203, opened 2015-03-28
refs #186 init script should still work when 'use_name' is defined

marji:
There are two changes:
1) use name and not cidfile in status function. That's because cidfile gets deleted from stop function, therefore it does not exist when the container is not running and the status function fails.

2) restore cidfile on docker start. We start by name anyway so it is safe to store the cid read via inspect.


garethr:
@marji thanks, I'd been meaning to find some time to look into this issue for a while. Much appreciated.

PR #205, opened 2015-04-02
Add ability to set log-level in docker configs

n0coast:
The default log level for docker is 'info' which can generate a lot of noise, just adding support for the -l|--log-level switch in the docker configs: /etc/conf.d/docker, /etc/default/docker, /etc/sysconfig/docker

From the manpage the usage is:

       -l, --log-level="debug|info|error|fatal""
         Set the logging level. Default is info.

This patch allows the optional docker::params::log_level to be set.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -21,14 +21,14 @@
 #   tcp://127.0.0.1:4243
 #   Defaults to undefined
 #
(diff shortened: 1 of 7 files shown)


n0coast:
This commit was a mess, rebasing and trying again

PR #206, opened 2015-04-02
adding support for log_level (-l) in docker config files

n0coast:
The default log level for docker is 'info' which can generate a lot of noise, just adding support for the -l|--log-level switch in the docker configs: /etc/conf.d/docker, /etc/default/docker, /etc/sysconfig/docker

From the manpage the usage is:

   -l, --log-level="debug|info|error|fatal""
     Set the logging level. Default is info.

This patch allows the optional docker::params::log_level to be set.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -25,6 +25,11 @@
 #   The unix socket to bind to. Defaults to
 #   unix:///var/run/docker.sock.
 #
+# [*log_level*]
+#   Set the logging level
+#   Defaults to undef: docker defaults to info if no value specified
+#   Valid values: degug, info, error, fatal
+#
 # [*use_upstream_package_source*]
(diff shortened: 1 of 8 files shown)


n0coast:
Sorry for the open/close... somehow the Gemfile.lock got updated, and I didn't want that bit committed.

garethr:
Given we know the valid values for this param it would be nice to put that into the code. Something in init.pp like:

```puppet
validate_re($log_level, '^(debug|info|error|fatal)$', 'log_level must be one of debug, info, error of fatal')
```

A quick unit test for this showing it fail for an invalid value would also be great.

Thanks for this change too, much appreciated.

n0coast:
Good ideas, implemented and build is passing.

garethr:
@n0coast Thanks for this addition, and for adding the additional tests.

PR #208, opened 2015-04-02
fix systemd template w.r.t. extra_parameters

jtriley:
The `<%= @image %>` tag should come immediately before the `<%= @command %>` otherwise docker will interpret any `extra_parameters` as the command argument and fail. The existing `/etc/init.d/docker-run.erb` template does this correctly. This fix makes both init.d and systemd templates consistent.


garethr:
@jtriley thanks, much appreciated.

PR #209, opened 2015-04-03
Add options for init script to clean .cid file

bchecketts:
The current init script writes a .cid file with the container id in /var/run/$prog.cid

If one wishes to pull down a new version of the container and start it up, this cannot be accomplished with the current init script alone since it re-uses the previous container id stored in the .cid file.

This pull request adds a 'clean' option to the init script which will remove the .cid file.  It also provides a cleanRestart option that will stop the container, remove the .cid file, and then start the container up again (with the latest image if if 'docker pull' has a new one available)

### DIFF
--- a/templates/etc/init.d/docker-run.erb
+++ b/templates/etc/init.d/docker-run.erb
@@ -123,6 +123,18 @@ stop() {
     fi
 }
 
+clean() {
+    if ! [ -f $cidfile ]; then
+        failure
+        echo
+        printf "$cidfile does not exist.\n"
+    else
+        rm $cidfile
+        retval=$?
+        return $retval
+    fi
+}
+
 case "$1" in
     start)
     start
@@ -141,11 +153,19 @@ case "$1" in
     stop
     start
     ;;
+    clean)
+    clean
+    ;;
+    cleanRestart)
+    stop
+    clean
+    start
+    ;;
     condrestart)
     [ -f /var/lock/subsys/$prog ] && restart || :
     ;;
     *)
-    echo "Usage: $0 [start|stop|status|reload|restart|probe]"
+    echo "Usage: $0 [start|stop|status|reload|restart|probe|clean|cleanRestart]"
     exit 1
     ;;
 esac


garethr:
I think this is a good idea. It would be nice to expose this to the `docker::run` type though.

Specifically, I think you could pass a suitably named toggle on `docker::run` to enable or disable cleanRestart, using the ability to pass a specific restart attribute value to the underlying service:
https://docs.puppetlabs.com/references/latest/type.html#service-attribute-restart

bchecketts:
The problem boils down to there being two desired behaviors:
 1) restart the container reusing the same docker image that it was created with
 2) restart the container with a new version of the docker image.

The current 'restart' command accomplishes use case 1, but there is not currently a way to accomplish 2 via the init script alone.  With two different desired behaviors, I believe there needs to be two different 'restart' options available.

In keeping with the style of command-line options, perhaps my proposed 'cleanRestart' option should be renamed 'clean-restart'

garethr:
@bchecketts thanks, as noted I do think this is a good idea. I'm also fine with the name in the init script. If we're happy with the terminology of clean-restart then I'd be looking for something like:

```puppet
docker::run { 'thing':
  ...
  clean-restart => true,
}
```

br0ch0n:
So, this ties into my need/desire to be able to upgrade docker containers and generally have the docker::run resource act similar to the puppet package resource.  Since lots of people treat docker containers as just another package type, we should be able to work with them accordingly.  i.e. changing the explicit image tag should ensure that the stated image tag is indeed running.  Right now, changing it updates the init script but doesn't reload the container with the new image.  I think this is one way we can accomplish that.  In my tests though, I also had to throw a potentially redundant 'docker rm' in there to get the behavior I want.  PR coming in a sec...

br0ch0n:
I've done https://github.com/garethr/garethr-docker/pull/215 for this

garethr:
Thanks both,  I've merged the related PR with this in.

Issue #211, opened 2015-04-08
systemd lacks ExecStart setting. Refusing.


jrenggli:
After enabling `pull_on_start` i get the following error message
```
[root@centos puppet]# puppet apply
...
Error: Could not start Service[docker-nginx-proxy]: Execution of '/bin/systemctl start docker-nginx-proxy' returned 6: Failed to issue method call: Unit docker-nginx-proxy.service failed to load: Invalid argument. See system logs and 'systemctl status docker-nginx-proxy.service' for details.
Error: /Stage[main]/Main/Docker::Run[nginx-proxy]/Service[docker-nginx-proxy]/ensure: change from stopped to running failed: Could not start Service[docker-nginx-proxy]: Execution of '/bin/systemctl start docker-nginx-proxy' returned 6: Failed to issue method call: Unit docker-nginx-proxy.service failed to load: Invalid argument. See system logs and 'systemctl status docker-nginx-proxy.service' for details.
Notice: Finished catalog run in 0.25 seconds
```

Digging in the log files...
```
[root@centos puppet]# /bin/systemctl start docker-nginx-proxy
Failed to issue method call: Unit docker-nginx-proxy.service failed to load: Invalid argument. See system logs and 'systemctl status docker-nginx-proxy.service' for details.

[root@centos puppet]# systemctl status docker-nginx-proxy.service
docker-nginx-proxy.service - Daemon for nginx-proxy
   Loaded: error (Reason: Invalid argument)
   Active: inactive (dead)

Apr 08 19:27:32 centos.dev systemd[1]: docker-nginx-proxy.service lacks ExecStart setting. Refusing.
Apr 08 19:29:22 centos.dev systemd[1]: docker-nginx-proxy.service lacks ExecStart setting. Refusing.
Apr 08 19:30:40 centos.dev systemd[1]: docker-nginx-proxy.service lacks ExecStart setting. Refusing.
Apr 08 19:30:54 centos.dev systemd[1]: docker-nginx-proxy.service lacks ExecStart setting. Refusing.
```

jrenggli:
It seems the problem ist, a missing new line in the file `/etc/systemd/system/docker-nginx-proxy.service`. Therefore the statement `ExecStart` isn't recognized
```
[Service]
...
ExecStartPre=-/usr/bin/docker kill nginx-proxy
ExecStartPre=-/usr/bin/docker rm nginx-proxy
[...]

PR #212, opened 2015-04-08
[BUGFIX] Add newline between ExecStartPre and ExecStart configuration

jrenggli:
Fix for #211


garethr:
@jrenggli thanks for reporting and then fixing that, much appreciated.

PR #213, opened 2015-04-09
Allow logging in and out of private registries that require authenticati...

pkill:
...on

These changes are offered as-is and are not hardened by rspec or acceptance tests.

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -1,0 +1,59 @@
+# == Class: docker
+#
+# Module to configure private docker registries from which to pull Docker images
+# If the registry does not require authentication, this module is not required.
+#
+# === Parameters
+# [*server*]
+#   The hostname and port of the private Docker registry. Ex: dockerreg:5000
+#
+# [*ensure*]
+#   Whether or not you want to login or logout of a repository
+#
+# [*username*]
+#   Username for authentication to private Docker registry.
+#   auth is not required.
+#
+# [*password*]
+#   Password for authentication to private Docker registry. Leave undef if
+#   auth is not required.
+#
+# [*email*]
+#   Email for registration to private Docker registry. Leave undef if
+#   auth is not required.
+#
+#
+define docker::registry(
+  $server    = $title,
+  $ensure    = 'present',
+  $username  = undef,
+  $password  = undef,
+  $email     = undef,
+) {
+  include docker::params
+
+  validate_re($ensure, '^(present|absent)$')
+
+  $docker_command = $docker::params::docker_command
+
+  if $ensure == 'present' {
+    if $username != undef and $password != undef and $email != undef {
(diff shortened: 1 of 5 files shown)


manifests/registry.pp
```diff
@@ -0,0 +1,59 @@
...
+  }
+  else {
+      $auth_cmd = "${docker_command} logout ${server}"
+  }
+
+  exec { "auth against ${server}":
+    command => $auth_cmd,
+    user    => 0,
+    path    => ['/bin', '/usr/bin'],
+    onlyif  => $auth_cmd,
```
garethr: Could you explain the rationale behind this?

I think what will happen here is that it will run the provided command, if that returns a 0 exit code it will run it again. I don't think this is required.

pkill: The exec will run a login or logout command only once and only if auth_cmd is set.

garethr: I've just checked the docs at https://docs.puppetlabs.com/references/latest/type.html#exec-attribute-onlyif and I don't think that's correct.

```
if this parameter is set, then this exec will only run if the command [in only_if] has an exit code of 0.
```

pkill: Yes, you're right. That line will be elided in the next push.

pkill:
@garethr added some puppet-rspec tests and a few fixes.

luxflux:
Would be cool to see this merged!

garethr:
@pkill thanks for this, and apologies for taking a while to get round to trying it out. Much appreciated.

luxflux:
Thanks for merging! Also for maintaining the awesome module! :beers:

dvorak:
@pkill What was the intent of allowing calling docker login without a username and password?  It seems like that always kicks it into interactive mode.

pkill:
@dvorak When dockerreg caches your creds, you don't have to supply credentials thereafter. This allows you to supply credentials outside of your puppet code if desired. Docker won't interactively ask you for your username and password at that point.

dvorak:
But at that point it's already cached them, so won't any further commands use them automatically?

dvorak:
The reason I ask is that I'm working on a patch to use augeas for putting the credentials directly into /root/.docker/config.json.  I think this is slightly gross, but using the docker login command there appears to be no way to prevent the exec from running every time.  Putting the credentials directly into the config file seems to be sufficient to make it work with any further commands.

pkill:
@dvorak perhaps your patch could use the onlyif parameter on the exec to check if the creds are cached already.

dvorak:
I was originally thinking that way, but I couldn't come up with a good way to do that.  Augeas has built in support for this sort of check, but only on augeas resources.  If I could assume jq or something similar was installed then I could probably pretty easily do it.

PR #214, opened 2015-04-09
Improve support for Debian 7

rji:
Prior to this commit, using this module on Debian 7 "Wheezy" with all
of the defaults required manually passing in the package name
(lxc-docker) because the case was not handled in params.pp. The upstream
repo is enabled by default, so I'd suspect that the least surprise to
users would be that the default package name was the same as it is in
the upstream repo: lxc-docker.

This commit modifies params.pp to include better defaults for Debian
Wheezy and modifies the test case accordingly.

For what it's worth, here's how we're using the Docker module at Puppet Labs:

```puppet
  class { '::docker':
    prerequired_packages => ['cgroupfs-mount'],
    package_name         => 'lxc-docker',
    service_name         => 'docker',
    docker_command       => 'docker',
  }
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,3 +82,6 @@ Or with Ubuntu 14.04 with:
 
     BEAKER_set=ubuntu-server-1404-x64 bundle exec rake beaker
 
+Or with Debian 7.8 with:
+
+    BEAKER_set=debian-78-x64 bundle exec rake beaker
--- a/Gemfile
+++ b/Gemfile
@@ -21,4 +21,5 @@ group :development do
   gem "guard-rake"
   gem "pry"
   gem "yard"
+  gem "vagrant-wrapper"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,8 +1,8 @@
 GIT
   remote: https://github.com/puppetlabs/beaker.git
-  revision: 2d25d06d20ee7e00d353e1df5c530f957e4c8cfa
+  revision: c68967e3d15674c1a39510222e14cfbaf6c59ad6
   specs:
-    beaker (2.8.0)
+    beaker (2.9.0)
       aws-sdk (~> 1.57)
       docker-api
       fission (~> 0.4)
@@ -20,7 +20,7 @@ GIT
 
 GIT
   remote: https://github.com/rodjek/rspec-puppet.git
-  revision: a9a837669cf6955279f02d1d9b524dc140b9d3e8
+  revision: 7ee8bd6ce2bf5eb8bea319ac22385ae0edde406e
   specs:
     rspec-puppet (2.0.1)
       rspec
@@ -29,21 +29,24 @@ GEM
   remote: http://rubygems.org/
   specs:
     CFPropertyList (2.3.1)
-    activesupport (3.2.21)
-      i18n (~> 0.6, >= 0.6.4)
-      multi_json (~> 1.0)
+    activesupport (4.2.1)
+      i18n (~> 0.7)
+      json (~> 1.7, >= 1.7.7)
+      minitest (~> 5.1)
+      thread_safe (~> 0.3, >= 0.3.4)
+      tzinfo (~> 1.1)
     addressable (2.3.8)
     autoparse (0.3.3)
       addressable (>= 2.3.1)
       extlib (>= 0.9.15)
       multi_json (>= 1.0.0)
-    aws-sdk (1.63.0)
-      aws-sdk-v1 (= 1.63.0)
-    aws-sdk-v1 (1.63.0)
+    aws-sdk (1.64.0)
+      aws-sdk-v1 (= 1.64.0)
+    aws-sdk-v1 (1.64.0)
       json (~> 1.4)
       nokogiri (>= 1.4.4)
     backports (3.6.4)
-    beaker-rspec (5.0.1)
+    beaker-rspec (5.0.2)
       beaker (~> 2.0)
       rspec
       serverspec (~> 2)
@@ -53,14 +56,14 @@ GEM
       timers (~> 4.0.0)
     coderay (1.1.0)
     diff-lcs (1.2.5)
-    docker-api (1.20.0)
+    docker-api (1.21.1)
       excon (>= 0.38.0)
       json
-    domain_name (0.5.23)
+    domain_name (0.5.24)
       unf (>= 0.0.5, < 1.0.0)
     ethon (0.7.3)
       ffi (>= 1.3.0)
-    excon (0.45.1)
+    excon (0.45.3)
     extlib (0.9.16)
     facter (1.7.6)
     faraday (0.9.1)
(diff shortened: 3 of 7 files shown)


garethr:
Cheers, Debian support has come in bits and bobs over the last few months. Could you confirm that with these changes Debian 7 should just work? If we I'll add it to the metadata, supported list on the repo and to the acceptance tests. Much appreciated.

rji:
The module has worked for me with the changes I mentioned above, _with one caveat_: Docker only "supports" Kernel >= 3.8, and Debian 7.x provides Kernel 3.2.x. The module configures the system, installs lxc-docker, and configures it all fine, the services start and the docker0 networking interface is up and working, but eventually when you try to run containers there will (probably) be some unfortunate consequences: https://github.com/docker/docker/issues/407%20kernel%20versions

_Note: this doesn't apply to EL 6.4 and newer, where the necessary commits have been backported to Kernel 2.6.39._

By enabling the wheezy-backports repo, you can install Kernel 3.16.x. But because of the way Puppet handles version strings for packages, and the way that Debian packaging ensures only the latest version in repositories, there isn't a _great_ way to puppetize that without bumping the Kernel version in the module (or a site's Hiera data) for every new kernel release (and you probably don't want `ensure => latest` on the kernel :smile:).

For what it's worth, on our infrastructure we have wheezy-backports available everywhere and this worked outside the manifest:
```
roger@cim-slave6-prod:~$ sudo puppet apply -e "package { 'linux-image-amd64': ensure => '3.16+63~bpo70+1', install_options => ['-t=wheezy-backports'] }"
Notice: Compiled catalog for cim-slave6-prod.delivery.puppetlabs.net in environment production in 0.03 seconds
Notice: /Stage[main]/Main/Package[linux-image-amd64]/ensure: ensure changed '3.2+46' to '3.16+63~bpo70+1'
Notice: Finished catalog run in 37.69 seconds
```

I'm still discussing this with our ops team, but so far I think we're leaning towards making Kernel >= 3.8 a prerequisite (outside of Puppet) for any systems that we intend to run Docker containers. I think we should proceed with making Debian Wheezy a supported platform for this module, if only because everything already works. If we do, I think it's worthwhile to perhaps include a "known issues" or "caveats" section in the README though.

rji:
If you'd like, I could also do the work here to ensure that we also use apt::backports and manage the kernel package on Debian Wheezy.

rji:
After further discussion with the ops team, we decided to manage everything with Puppet, at the risk of having to bump the kernel version whenever more patches were applied to Kernel 3.16 in wheezy-backports. Here's the code we're using to do that:

```puppet
  # Debian Wheezy ships with Kernel 3.2.0, and Docker requires Kernel >= 3.8.
  # Backports is already enabled, so let's upgrade the Kernel.
  #
  # For more info, see
  # https://docs.docker.com/installation/debian/#debian-wheezystable-7x-64-bit
  if $::operatingsystem == 'Debian' and $::lsbmajdistrelease == '7' {
    $kernel_version     = '3.16.0'
    $kernel_apt_version = '3.16+63~bpo70+1'

    package { 'linux-image-amd64':
      ensure          => $kernel_apt_version,
      install_options => ['-t=wheezy-backports'],
    }

    if $::kernelversion != $kernel_version {
      notify { "Please reboot the system to load Kernel ${kernel_version}": }
    }
  }
```

It would be trivial for me to add this to the PR and change the scope to "Better support for Debian 7", since you already have the logic around `manage_kernel`.

garethr:
Adding this to the PR would be great. Thanks. I also need to investigate the `reboot` module now that has some support for Linux I believe.

rji:
Hey @garethr, I haven't forgotten about this. I was working on wrapping up some other sprint work before revisiting my patch here. I'll try to get you something solid by Monday or Tuesday of next week.

rji:
Hrm, it looks like there's an issue when running acceptance. At first glance, it _looks_ like an issue with resource containment and the puppetlabs-apt module.

garethr:
@rji could you post a gist of the output?

It's likely worth rebasing from master as the issue may have been resolved in https://github.com/garethr/garethr-docker/commit/55805ca48577ca422dfe83de0023a94655be040c.

rji:
Hey @garethr - It looks like this wasn't related to 55805ca after all. The Apt::Backports class appears to be creating the wheezy-backports config fine on the SUT, but it doesn't appear to be usable. I'm still investigating.

```
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold -t=wheezy-backports install apt-transport-https' returned 100: Reading package lists...
E: The value 'wheezy-backports' is invalid for APT::Default-Release as such a release is not available in the sources
Error: /Stage[main]/Docker::Install/Package[apt-transport-https]/ensure: change from purged to present failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold -t=wheezy-backports install apt-transport-https' returned 100: Reading package lists...
E: The value 'wheezy-backports' is invalid for APT::Default-Release as such a release is not available in the sources
```

rji:
Actually, it looks like a red herring: wheezy-backports is fine. Instead, Package[apt-transport-https] isn't coming early enough in the manifest, and Exec[apt_update] fails when trying to fetch the Docker repo since it uses https (at https://github.com/garethr/garethr-docker/blob/master/manifests/params.pp#L53). Because Exec[apt_update] failed, wheezy-backports isn't usable.

```
Notice: /Stage[main]/Apt::Backports/Apt::Pin[backports]/File[backports.pref]/ensure: created
Notice: /Stage[main]/Docker::Install/Apt::Source[docker]/Apt::Pin[docker]/File[docker.pref]/ensure: created
Notice: /Stage[main]/Docker::Install/Apt::Source[docker]/File[docker.list]/ensure: created
Info: /Stage[main]/Docker::Install/Apt::Source[docker]/File[docker.list]: Scheduling refresh of Exec[apt_update]
Info: /Stage[main]/Docker::Install/Apt::Source[docker]/File[docker.list]: Scheduling refresh of Exec[Required packages: 'debian-keyring debian-archive-keyring' for docker]
Notice: /Stage[main]/Apt::Backports/Apt::Source[backports]/File[backports.list]/ensure: created
Info: /Stage[main]/Apt::Backports/Apt::Source[backports]/File[backports.list]: Scheduling refresh of Exec[apt_update]
Notice: /Stage[main]/Apt/File[/etc/apt/apt.conf.d/15update-stamp]/ensure: defined content as '{md5}4355b3e7824866a503fc221621fc65ba'
Notice: /Stage[main]/Docker::Install/Apt::Source[docker]/Exec[Required packages: 'debian-keyring debian-archive-keyring' for docker]: Triggered 'refresh' from 1 events
```

```
Notice: /Stage[main]/Apt::Update/Exec[apt_update]/returns: E: The method driver /usr/lib/apt/methods/https could not be found.
Error: /Stage[main]/Apt::Update/Exec[apt_update]: Failed to call refresh: /usr/bin/apt-get update returned 100 instead of one of [0]
Error: /Stage[main]/Apt::Update/Exec[apt_update]: /usr/bin/apt-get update returned 100 instead of one of [0]
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold -t=wheezy-backports install apt-transport-https' returned 100: Reading package lists...
```

rji:
Closing stale PR.

Issue #93, opened 2014-09-12
pull latest container


hackzilla:
I'm trying to figure out the best way to upgrade the docker containers using puppet.

Currently trying:
```
	docker::image { 'kohana-remove':
	  ensure  => 'absent',
	  image   => 'docker.ukwm.local:5000/ukwm/container',
	  require => Class['ukwm::packages::docker'],
	}
    ->
	docker::image { 'kohana-install':
	  ensure  => 'latest',
	  image   => 'docker.ukwm.local:5000/ukwm/container',
	  require => Class['ukwm::packages::docker'],
	}
    ->
	docker::run { 'kohana':
	  image           => 'docker.ukwm.local:5000/ukwm/container',
	  ports           => ['81:80', '2221:22'],
	  use_name        => true,
	  restart_service => true,
	  privileged      => false,
	}
```

Which outputs:
```
Error: docker rmi docker.ukwm.local:5000/ukwm/container returned 1 instead of one of [0]
Error: /Stage[main]/Stats::Backend::Kohana/Docker::Image[kohana-remove]/Exec[docker rmi docker.ukwm.local:5000/ukwm/container]/returns: change from notrun to 0 failed: docker rmi docker.ukwm.local:5000/ukwm/container returned 1 instead of one of [0]
Warning: /Stage[main]/Ukwm::Backend::Kohana/Docker::Image[kohana-install]/Exec[docker pull docker.ukwm.local:5000/ukwm/stats-kohana]: Skipping because of failed dependencies
Warning: /Stage[main]/Ukwm::Backend::Kohana/Docker::Run[kohana]/File[/etc/init/docker-kohana.conf]: Skipping because of failed dependencies
Warning: /Stage[main]/Ukwm::Backend::Kohana/Docker::Run[kohana]/Service[docker-kohana]: Skipping because of failed dependencies
```

So maybe an option of force ```docker rm -f <container>```

What would be nicer, is if could notify run that the image has been update. Then it could stop, remove, and restart container.

This would be the command I'd use.
```
	docker::image { 'kohana-install':
	  ensure  => 'latest',
	  image   => 'docker.ukwm.local:5000/ukwm/container',
	  require => Class['ukwm::packages::docker'],
	}

	docker::run { 'kohana':
	  image           => 'docker.ukwm.local:5000/ukwm/container',
	  ports           => ['81:80', '2221:22'],
	  use_name        => true,
	  restart_service => true,
	  privileged      => false,
	}
```

[...]

PR #215, opened 2015-04-10
Issue #93 allow versioning of containers

br0ch0n:
This builds on @bchecketts ' code from https://github.com/garethr/garethr-docker/pull/209, changing the init script and adding a param, clean_restart.
This allows one to change the image tag in docker::run from say redis:2.6 to redis:2.8.14 and have the old container destroyed and a new one running the new image put in its place.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -32,6 +32,7 @@ define docker::run(
   $socket_connect = [],
   $hostentries = [],
   $restart = undef,
+  $clean_restart = false,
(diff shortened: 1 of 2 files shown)


garethr:
Thanks both @br0ch0n and @bchecketts for this change. I'm going to merge this now, but it's also worth following https://github.com/garethr/garethr-docker/pull/216 too which might change the need for this in the future.

PR #216, opened 2015-04-11
Make the debian init script not use the cid if use_name is true

solarkennedy:
In the systemd init definition in this module, we use the "name" and assume it is the *only* version of that docker container running. We even docker rm as a pre-start.

This change ports that functionality into the general debian init script.

The rationale here is that if we `use_name`, we only ever want one version of the container running ever.

This is especially importance if we have a restart policy on a container. If we ask for `--restart=always`, then docker will restart the container, the cid changes, and this init script is ignorant of that change. (leading to multiple stacked running containers)

This code change makes it safe to use a restart policy, and makes it so --use-name really does make it use the name everywhere (which is always constant) instead of the cid (which can get stale, and is in general flaky)

The good news is: using the name actually makes the output bash version of this script much simpler. If you ever want to ever make the init.d change exactly compatible with the systemd version, we can get rid of all this cid junk entirely. Heck, we already know the "name", it is the name of the puppet resource title. Shouldn't any puppet-run docker always be referenced by name?

cc @mrtyler @EvanKrall

### DIFF
--- a/spec/defines/run_spec.rb
+++ b/spec/defines/run_spec.rb
@@ -69,6 +69,7 @@ require 'spec_helper'
     context 'when `use_name` is true' do
       let(:params) { {'command' => 'command', 'image' => 'base', 'use_name' => true } }
       it { should contain_file(initscript).with_content(/ --name sample /) }
+      it { should contain_file(initscript).with_content(/docker rm sample/) }
     end
 
     context 'when stopping the service' do
--- a/templates/etc/init.d/docker-run.erb
+++ b/templates/etc/init.d/docker-run.erb
@@ -58,6 +58,14 @@ fi
 start() {
     [ -x $docker ] || exit 5
 
+    <% if @use_name -%>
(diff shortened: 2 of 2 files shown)


solarkennedy:
Eh, hold on, this isn't quite right. We can't run "docker start" on a named container that doesn't exist...

solarkennedy:
@garethr all this cid nonsense, stale cid, etc makes this init script very complicated.

Would you accept a PR to *always* use the name of the container like the systemd version?

This would greatly simplify the init script/tests/etc. Additionally there would *Never* be any stale cid files or double-running containers.

It also greatly simplifies the status call, as it is either running, or not. And it makes it compatible with restart policies.

garethr:
@solarkennedy short answer: yes.

I simplified this  a little back and have the same opinion about the cid file flakiness now I've looked into it.

solarkennedy:
Yea, like #93 is just more hacks :(
I'll work on this soon as it is causing stacked containers (multiple docker runs with the same image, different ephemeral name)

solarkennedy:
I'm going to discard this in favor of a new PR

PR #220, opened 2015-04-14
Acceptance tests

jonnytpuppet:
Acceptance tests


spec/acceptance/docker_basic_crud_spec.rb
```diff
@@ -0,0 +1,142 @@
+require 'spec_helper_acceptance'
+
+describe 'docker CRUD tests' do
```
garethr: This should describe what you're testing, so in this case the docker Puppet class.

spec/acceptance/docker_basic_crud_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+          }
+      EOS
+
+      apply_manifest(pp, :catch_failures => true)
+      unless fact('selinux') == 'true'
+        apply_manifest(pp, :catch_changes => true)
+      end
+    end
+
+    it 'should contain the output' do
```
garethr: This should be explicit, it should be useful without the body of the test

spec/acceptance/docker_basic_crud_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      shell('docker ps') do |r|
+        expect(r.stdout).to match(/ubuntu:14.04\s+"init".+5555\/tcp\, 0\.0\.0.0\:\d+\-\>4444\/tcp\s+demo1/)
+      end
+      shell('docker images') do |r|
```
garethr: This should probably be a separate test to the one above, ditto the one below.

spec/acceptance/docker_basic_crud_spec.rb
```diff
@@ -0,0 +1,142 @@
+require 'spec_helper_acceptance'
+
+describe 'docker CRUD tests' do
+  context 'docker add' do
+    it 'installs docker, pulls images and runs container' do
```
garethr: This should probably be a before block for this context. It's really about setting up the assertions I think.

spec/acceptance/docker_basic_crud_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    end
+
+    it 'should contain the output' do
+      shell('docker ps') do |r|
+        expect(r.stdout).to match(/ubuntu:14.04\s+"init".+5555\/tcp\, 0\.0\.0.0\:\d+\-\>4444\/tcp\s+demo1/)
+      end
+      shell('docker images') do |r|
+        expect(r.stdout).to match(/ubuntu\s+14.04/)
+      end
+      shell('netstat -tulp | grep 4243') do |r|
```
garethr: The test text can describe what this test is doing.

spec/acceptance/docker_basic_crud_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+          ports => ['1111'],
+          expose => ['2222'],
+          memory_limit => '200m',
+          use_name => true,
+          }
+      EOS
+
+      apply_manifest(pp, :catch_failures => true)
+      apply_manifest(pp2, :catch_failures => true)
+      unless fact('selinux') == 'true'
```
garethr: What's the rationale for this clause? It's not immediately obvious from the code so might warrant a individual test or a comment.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,598 @@
+require 'spec_helper_acceptance'
+
+describe 'docker full acceptance tests section 1' do
+  context 'test 1.1 - Install and configure Docker and Docker Daemon' do
```
garethr: I'm not sure I like the `test 1.1` portion, I don't think keeping references like this will be particularly sustainable.

spec/acceptance/docker_image_crud_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'spec_helper_acceptance'
```
garethr: Slightly rewritten version https://gist.github.com/garethr/c6b8f5398d8b9e69c6de

spec/acceptance/nodesets/ubuntu-1404-x64-vcloud.yml
```diff
@@ -0,0 +1,16 @@
+HOSTS:
```
garethr: I think we should probably keep these in the Puppet Labs fork, along with other vendor modifications.

jonnytpuppet:
This can now be merged I believe. I've added a Temp hack to require the latest device-mapper for RHEL7/CentOS7 and also install docker from the rhel-extras repo for RHEL7.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+              require => Package['device-mapper']
+            }
+      EOS
+      apply_manifest(pp, :catch_failures => true)
+  end
+end
+
+describe 'docker full acceptance tests section 1 - testing class {\'docker\':}' do
+  context 'test 1.1 - Install and configure Docker and Docker Daemon' do
+    it 'applies the manifest' do
```
garethr: This should be a before block for the two tests.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
+require 'spec_helper_acceptance'
+
+describe 'Temp OS hacks' do
```
garethr: One of these installs docker and the other doesn't? I think this assumes the other test cases are run on the same machine as these? This should be a before block on a global context if that's the intent. Otherwise (I think) other tests will fail if you run them individually.

spec/acceptance/docker_full_spec.rb
jonnytpuppet: I'm sorry I don't understand?

The tests are passing clean on RHEL7. CentOS7 (after quickly enabling a normal mirror) and Ubuntu 14.04 consistently

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+              install_options => ['--enablerepo=rhel7-extras'],
+              ensure => latest,
+              require => Package['device-mapper']
+            }
+      EOS
+      apply_manifest(pp, :catch_failures => true)
+  end
+end
+
+describe 'docker full acceptance tests section 1 - testing class {\'docker\':}' do
```
garethr: I'd rather this described what you are testing. So just ' the Puppet docker class' or similar.

spec/acceptance/docker_full_spec.rb
jonnytpuppet: These tests are written for the 3 supported OS.

We only need to install docker manually to enable the extras repo for RHEL.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+        class { 'docker': }
+      EOS
+
+      apply_manifest(pp, :catch_failures => true)
+      unless fact('selinux') == 'true'
+        apply_manifest(pp, :catch_changes => true)
+      end
+    end
+
+    it ': ps -aux should be running docker process' do
```
garethr: This shouldn't repeat the implementation details, so something like 'should successfully start a docker daemon'

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+      shell('ps -aux | grep docker') do |r|
+        expect(r.stdout).to match(/\/usr\/bin\/docker/)
+      end
+    end
+    it ': docker ps should output table headers' do
```
garethr: This is quite brittle, given that the output of ps can and does change. I'd maybe just change to to check for a zero exit code.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+      shell('docker ps') do |r|
+        expect(r.stdout).to match(/CONTAINER ID\s+IMAGE\s+COMMAND\s+CREATED\s+STATUS\s+PORTS\s+NAMES/)
+      end
+    end
+  end 
+
+  context 'test 1.3a - Bound to a particular TCP socket' do
```
garethr: change to 'when providing a TCP address to bind to'

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+        }
+      EOS
+
+      apply_manifest(pp, :catch_failures => true)
+      unless fact('selinux') == 'true'
+        apply_manifest(pp, :catch_changes => true)
+      end
+    end
+
+    it ': netstat should show docker listening' do
```
garethr: remove netsta (ie. the implementation details.) I'll assume that for other tests as well.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+      end
+    end
+    it ': docker ps should output the table headers' do
+      shell('docker ps') do |r|
+        expect(r.stdout).to match(/CONTAINER ID\s+IMAGE\s+COMMAND\s+CREATED\s+STATUS\s+PORTS\s+NAMES/)
+      end
+    end
+  end  
+
+  context 'test 1.3b - Bound to a particular unix socket' do
```
garethr: I don't like the test 1.3b part, it feels like it could refer to something that's not part of the repo. If you could remove from all the tests.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+    # Delete all existing images
+    shell('docker rmi $(docker images -q) || true')
+    # Check to make sure no images are present
+    shell('docker images | wc -l') do |r|
+      expect(r.stdout).to match(/^0|1$/)
+    end
+  end
+
+  context 'test 2.1 - Download Image' do
+    it 'runs test' do
```
garethr: describe the test. So in full this should probably be:

    docker::image should successfully download an image from the Docker Hub

You don't actually need the context block here.

spec/acceptance/docker_full_spec.rb
```diff
@@ -0,0 +1,643 @@
...
+      end
+
+      shell('docker images') do |r|
+        expect(r.stdout).to_not match(/ubuntu/)
+      end
+    end
+  end
+end
+
+[true,false].each do |use_name|
```
garethr: As of https://github.com/garethr/garethr-docker/pull/222 I don't think this is now required.

PR #221, opened 2015-04-15
Adding support for --selinux-enabled=true|false option

n0coast:
spec tests created and passing


garethr:
@n0coast thanks, this looks good. Much appreciated.

PR #222, opened 2015-04-15
Make the debian init script always use the title as the name of the docker container

solarkennedy:
We already have a name for the docker::run, it is the title.

This PR makes all docker::run's use the title as the primary key, instead of the cid file.

This greatly simplifies the (debian) init script, makes it compatible with restart policies, and makes it much more obvious to the user when they run docker-ps themselves. (and don't have to remember to cleanup cid files if they manually kill something)

Obsoletes #190
Related to #209, #203

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -178,8 +178,9 @@ define docker::run(
     if $initscript == "/etc/init.d/docker-${sanitised_title}" {
       # This exec sequence will ensure the old-style CID container is stopped
       # before we replace the init script with the new-style.
-      exec { "/etc/init.d/docker-${sanitised_title} stop":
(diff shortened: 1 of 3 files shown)


solarkennedy:
I also have code to do the transition help (from cid to name) that I have in our profile_docker, but if you want I can make a separate PR for that if you want it to be in here.

Additionally if accepted I'll do a second pr to remove the use_name parameter entirely.

solarkennedy:
I think it is a bit irresponsible to ship such a breaking change without code to help the transition, so I've included it in this PR.

garethr:
@solarkennedy thanks again for this, I've just made some fixes due to upstream Docker changes in RedHat so I'll hopefully be releasing a new version to the Forge soon.

Issue #130, opened 2014-10-13
[Feature Request] docker::stop  / docker::rm


farion:
It should be possible to stop and remove containers somehow Basically the opposite of docker::run. Otherwise it is not possible to get rid of containers you don't need anymore. 

Perhaps something like this:

// Will stop all containers with that image (and tag)
docker::stop { 'helloworld':
  image   => 'base:tag'
}

// Will remove all containers with that image (and tag)
docker::remove { 'helloworld':
  image   => 'base:tag'
}

MichaelMackus:
:+1:

garethr:
As of https://github.com/garethr/garethr-docker/pull/403 the module now supports making containers absent. Thanks for the report, and apologies it took a while to add the capability.

PR #223, opened 2015-04-16
Removed restart setting from systemd erb file.

jonnytpuppet:
Removed restart setting from systemd erb file.

### DIFF
--- a/templates/etc/systemd/system/docker-run.erb
+++ b/templates/etc/systemd/system/docker-run.erb
@@ -8,7 +8,6 @@ Requires=docker.service<% if @depends %><% @sanitised_depends_array.each do |d| 
 
 [Service]
 <%- if @username -%>User=<%= @username %><%- end -%>
-Restart=on-failure
 StartLimitInterval=20
 StartLimitBurst=5
 TimeoutStartSec=0


garethr:
I'm not sure this the right think.

This setting appears useful to me, specifically allowing systemd to ensure a container is running all the time. Without this you're relying on an external process (like monit), or it not crashing, or Puppet once every half hour (too long)?

It would be useful to explain the rationale for why this would be a good idea?

jonnytpuppet:
The default behaviour in docker is restart=no. (I.e. by removing the setting, this is the restart setting that is applied)

Additionally, the docker module does not work properly when restart is set to anything else other than restart=no (or by leaving the restart setting out). 

The things that break are:
use_name
stopping a container
docker rm command line tool

Removing restart also keeps things consistent with how the init.d scripts work for debian based OSes.

garethr:
1. I think we're going to kill `use_name` soon so I'm not concerned with the difference between the init scripts at present
2. That's not supported by the module at all at present, it's being tracked in https://github.com/garethr/garethr-docker/issues/130
3. If you need to remove a container, first stop the systemd service. This should prevent systemd restarting the container when you remove it.
4. I think the fact that restarts aren't nicely handled by sysvinit is a good reason to use systemd, rather than a reason to remove useful functionality.

Sorry, I don't see this being merged. The individual issues are just that, individual. And the suggested fix is both backwards incompatible and makes the module worse.

PR #224, opened 2015-04-18
Remove unused references to use_name

solarkennedy:
The removes all references to use_name.
Notably absent here is changes to the systemd script, which never used this parameter in the first place.
The normal init script matches this behavior, so this parameter is cruft.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,9 +187,9 @@ docker::run { 'helloworld':
 }
 ```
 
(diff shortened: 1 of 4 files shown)


garethr:
My plan for this is to ship a version to the Forge soon, with a deprecation notice in place for this. I'll let that sit a little and then merge this.

PR #226, opened 2015-04-18
Removed 'restart' init script workaround

solarkennedy:
This code looks like it was a workaround for the misbehaving cid behavior with restart policies.
It is no longer needed. No tests need changing as this code was not under test before.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -71,71 +71,51 @@ define docker::run(
   $depends_array = any2array($depends)
 
   $docker_run_flags = docker_run_flags({
-    cpuset => any2array($cpuset),
-    detach => $valid_detach,
+    cpuset          => any2array($cpuset),
+    detach          => $valid_detach,
     disable_network => $disable_network,
-    dns => any2array($dns),
-    dns_search => any2array($dns_search),
-    env => any2array($env),
-    expose => any2array($expose),
-    hostname => $hostname,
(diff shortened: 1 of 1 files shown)


solarkennedy:
Er, hold on, need to rebase this off your master.

garethr:
That's not a workaround, but a way of using the built-in docker process manager instead of the init scripts. I'm not 100% on the interface as yet, although it makes sense from a certain point of view it's not very explicit.

Basically, at present, if you set a restart policy for docker, docker will manage the container instead of handing that off to the init system.

solarkennedy:
The restart policy is a good thing, but it doesn't need a special invocation / exec.

It can be in the plain init script. It *already* does this because of the named containers. 
https://github.com/garethr/garethr-docker/blob/master/templates/etc/init.d/docker-run.erb#L91
(stop/start etc)

We (Yelp) were emulating this before with using `extra_params => [ 'restart always']`.
Because even if you use a restart policy, you *still* want an init script, and it is much easier for puppet to control using a `service` than the execs.

garethr:
> you still want an init script

Actually, that's been a previous feature request, where people don't want the init involved and just want to hand this off to Docker. 

This should however be more explicit. It works but it's a little spooky and not well tested as yet.

solarkennedy:
Ok, I'll add some docs instead.

PR #227, opened 2015-04-18
Added docs explaining what the restart parameter on docker::run really does

solarkennedy:
While I personally think that the way that `restart` does things is ill-advised, I understand that some people want it to work like this.

This adds docs to help other people know how they can get a restart policy and have puppet continue to manage the container via the normal init script mechanism.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -1,6 +1,27 @@
 # == Define: docker:run
 #
-# A define which managed docker container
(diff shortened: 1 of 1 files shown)


garethr:
@solarkennedy thanks again. I definitely owe you a beverage of your choice next time your in the UK or I make it to SF.

PR #231, opened 2015-04-20
Added known issues to README.

jonnytpuppet:
Added known issues to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,4 +255,32 @@ docker::exec { 'helloworld-uptime':
   tty       => true,
 }
 ```
+## Known Issues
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -255,4 +255,32 @@ docker::exec { 'helloworld-uptime':
...
+```
+
+Remember to add the appropriate metaparameters (before or require) for your environment to ensure that device-mapper is installed before the docker class is executed.
+
+RHEL7 also has the above issue. 
+
+Additionally, RHEL7 needs the rhel7-extras repo enabled in order to install docker. You can install docker using the following example manifest:
+
+```puppet
+package {'docker':
```
garethr: This will result in a duplicate resource error if used with the Puppet module.

PR #234, opened 2015-04-21
(maint) Update sudo calling conventions

justinstoller:
For redhat based systems we were failing because the before(:all) was
calling sudo without a pty. Several ServerSpec examples were also
calling sudo directly instead of passing {:sudo => true} into the
describe. I don't know why this was passing on Debian.

This updates the command to tell Beaker to allocate a pty and use
ServerSpec's sudo facility.


justinstoller:
Passed locally testing on RHEL 7 and Trusty.

Iristyle:
:+1:

PR #236, opened 2015-04-23
Added support for automatically enabling the required RHEL7, OL7 and SL7 repos

jonnytpuppet:
Added support for automatically enabling the required RHEL7, OL7 and SL7 repos when installing docker.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -107,9 +107,17 @@ class docker::install {
   }
 
   if $docker::manage_package {
-    package { 'docker':
-      ensure => $docker::ensure,
-      name   => $dockerpackage,
+    if $docker::repo_opt {
+      package { 'docker':
+        ensure          => $docker::ensure,
+        name            => $dockerpackage,
+        install_options => $docker::repo_opt,
+      }
+    } else {
+        package { 'docker':
+          ensure => $docker::ensure,
+          name   => $dockerpackage,
+        }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


manifests/install.pp
```diff
@@ -107,9 +107,17 @@
...
 
   if $docker::manage_package {
-    package { 'docker':
-      ensure => $docker::ensure,
-      name   => $dockerpackage,
+    if $docker::repo_opt {
+      package { 'docker':
+          ensure          => $docker::ensure,
+          name            => $dockerpackage,
+          install_options => $docker::repo_opt,
```
garethr: Isn't that going to be in the params scope? ie. `$docker::params::repo_opt`?

manifests/install.pp
```diff
@@ -107,9 +107,17 @@
   }
 
   if $docker::manage_package {
-    package { 'docker':
-      ensure => $docker::ensure,
-      name   => $dockerpackage,
+    if $docker::repo_opt {
+      package { 'docker':
+          ensure          => $docker::ensure,
```
garethr: Whitespace issue for this package resource

manifests/install.pp
jonnytpuppet: They both work as docker inherits docker::params. Which is preferred?

garethr:
Could you add a simple unit test to validate this is going the right thing? ie. install_options isn;t set of some flavour of Debian and is set correctly for some flavour of RedHat.

PR #237, opened 2015-04-24
Convergent docker image

solarkennedy:
Related: https://github.com/Yelp/garethr-docker/pull/2

This adds a wrapper for docker pull to allow puppet to know when an update has been applied, so it can correctly converge and refresh other resources.

The addresses the problem of docker::image { 'foo': ensure => latest } never converging.
It also solves the problem of

```puppet
docker::image { 'foo': ensure => latest } ~>
docker::run { 'foo': }
```
always restarting the image. (because ensuring latest meant that it *always* refreshed the docker::run)

Additionally integration tests are in place to prevent this resource from ever non-converging in the future.


Gemfile
```diff
@@ -21,4 +21,5 @@ group :development do
   gem "guard-rake"
   gem "pry"
   gem "yard"
+  gem "vagrant-wrapper"
```
garethr: I've had problems previously with vagrant-wrapper, and had support issues with other modules. This shouldn't be required.

files/update_docker_image.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+#
+# Pulls a docker image.
+# Returns 0 if there a change.
+# Returns 1 if there is no change.
+# Returns 3 if something when wrong.
+#
+DOCKER_IMAGE=$1
+
+BEFORE=$(docker inspect --format='{{.Config.Image}}' ${DOCKER_IMAGE} 2>/dev/null)
```
garethr: The previous implementation allowed overriding the binary with `docker_command` but this script doesn't do that.

solarkennedy:
I've rebased and addressed feedback but I need to fix up some tests and double check the acceptance tests.

solarkennedy:
Now, previously I ran the acceptance tests (`bundle exec rake beaker`) but now when I do that, this is the error I get:

    /home/kyle/Projects/garethr-docker/vendor/bundle/ruby/2.1.0/gems/beaker-2.32.1/lib/beaker/hypervisor/vagrant.rb:196:in `block (2 levels) in vagrant_cmd': Failed to exec 'vagrant destroy --force'. Error was /usr/lib/ruby/vendor_ruby/bundler/rubygems_integration.rb:292:in `block in replace_gem': vagrant-wrapper is not part of the bundle. Add it to Gemfile. (Gem::LoadError) (RuntimeError)

Makes me... want to but `vagrant-wrapper` in the Gemfile?

garethr:
@solarkennedy I've had this problem before. From memory I think this means that rather than vagrant being on your path, you have a binary called vagrant put their by vagrant-wrapper. Try `which vagrant` and then cat the file, my guess is removing that will resolve the problem.

Vagrant wrapper seemed like a good idea, but it basically needs to be used by everyone everywhere to not cause problems.

solarkennedy:
@garethr gotcha, that was it!

I've updated things, here was the (truncated) output of the beaker tests:
spec/acceptance/docker_full_spec.rb

all passed.

PR #241, opened 2015-04-24
Correcting commenting level in docker-run template

ChrisCrewdson:
for dependency based boot sequencing.

I don't think a proper test can be written for this.


garethr:
@ChrisCrewdson ah, good spot. Thanks for the fix.

PR #242, opened 2015-04-24
Added else case for repo_opt

jonnytpuppet:
Added else case for repo_opt

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -53,6 +53,7 @@ class docker::params {
       $package_source_location     = 'https://get.docker.io/ubuntu'
       $use_upstream_package_source = true
       $detach_service_in_init = true
+      $repo_opt = undef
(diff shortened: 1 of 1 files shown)


garethr:
@jonnytpuppet  Does the problem that this is aimed to fix also occur for non-RedHat distros? They won't have `$repo_opt` set?

jonnytpuppet:
Yep I believe so. I added the else clause twice. This PR fixes issues with future parser.

jonnytpuppet:
Actually wait, you're correct, I need to define repo_opt in the other cases. I'll update the PR.

garethr:
@jonnytpuppet thanks. It would be useful to describe the problem, and if possible add a test case, in the title and description of the issue - rather than describing the change. I can more easily review things if I have the context.

PR #245, opened 2015-04-29
Reduced the APT Key to 8 chars to pass validation

Gingonic:
On Ubuntu 14.04 LTS using the module with version 4.0.2 I get : 

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: validate_re(): "36A1D7869245C8950F966E92D8576A8BA88D21E9" does not match ["\\A(0x)?[0-9a-fA-F]{8}\\Z", "\\A(0x)?[0-9a-fA-F]{16}\\Z"] at /etc/puppet/modules/apt/manifests/key.pp:60 on node mma01
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

It seems the key is validated on 8 or 16 chars by the APT module. I simply reduced the key in the install class to meet this requirements

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -25,7 +25,7 @@ class docker::install {
           release           => 'docker',
           repos             => 'main',
           required_packages => 'debian-keyring debian-archive-keyring',
-          key               => '36A1D7869245C8950F966E92D8576A8BA88D21E9',
+          key               => '36A1D71E',
           key_source        => 'https://get.docker.io/gpg',
           pin               => '10',
           include_src       => false,


garethr:
@Gingonic ah, this may be an issue with the metadata around the APT module. What version of the APT module are you using?

1.8.x definitely allows for the full 40 character: https://github.com/puppetlabs/puppetlabs-apt/blob/1.8.x/manifests/key.pp#L62

Gingonic:
Seems I'm on 1.7.0. So upgrading would fix it too. Then it becomes a backward compatibility issue

Gingonic:
On 1.8 a short key actually produces a warning : 

```
Warning: /Apt_key[Add key: E56151BF from Apt::Source mesosphere]: The id should be a full fingerprint (40 characters), see README.
```  

So I guess my merge request should be discarded or moved up to a bit more refactoring to support the old APT

garethr:
Thanks for investigating. I think the correct thing to do here is to update the metadata to indicate the module requires at least 1.8. The PR for that I think referenced the deprecation warning, but at the time I didn't realise this wasn't backwards compatible.

stupied4ever:
It will not be merged ?

garethr:
@stupied4ever no, as mentioned the metadata should stipulate at least puppetlabs-apt 1.8. I've just not got round to updating that yet alas.

garethr:
I've updated the metadata in #311 which should be merged in the next week or so. This stipulates the minimum apt module version of 1.8. Thanks all.

PR #251, opened 2015-05-13
Add ability to use environment variable files

mweigel:
This adds the ability to specify an array of files containing environment variables using Docker's '--env-file' option when running a container. The actual files containing the environment variables are assumed to exist.

docker::run { 'foo':
  env_file => ['/etc/foo.env', '/etc/bar.env']
}


lcarstensen:
+1.  I need to pass in environment variable contents with spaces in it (see https://gist.github.com/lcarstensen/0bae93f196f06232074f for use case) and where that fails today this  PR also addresses that use case.

garethr:
@mweigel thanks for this, a really useful addition. and thanks to @lcarstensen for testing it out. Much appreciated.

PR #253, opened 2015-05-14
added DOCKER CERT PATH into template

ed-mt:
to allow the overriding of the CERT_PATH


garethr:
This probably also wants to be updated in https://github.com/ed-mt/garethr-docker/blob/docker_cert_path/templates/etc/sysconfig/docker.rhel7.erb and https://github.com/ed-mt/garethr-docker/blob/docker_cert_path/templates/etc/default/docker.erb I think?

This might also be best to set default to undef and apply the relevant logic to the template to only set this if a user passes a value (which is the default behaviour for the docker CLI). Otherwise if the default changes we have to maintain it here too.

PR #254, opened 2015-05-22
Added Docker-compose

scotty-c:
Hey Gareth,

Added docker-compose  !!!


Modulefile
```diff
@@ -0,0 +1,12 @@
+name    'garethr-docker'
```
garethr: Modulefile is no longer required, it's been superseded by the metadata.json file

README.md
```diff
@@ -257,4 +257,21 @@ docker::exec { 'helloworld-uptime':
...
+  ensure    => file,
+  content   => template('helloworld/docker-compose.yml.erb'), 
+  }
+```
+Then will build the containers with the following statement
+
+```puppet
+docker_compose {'Hello World':
+  source   => '/root',  
+  require  => File['/root/docker-compose.yml']
```
garethr: Rather than always require people to do this manually it might be nicer to do this using autorequire in the type (eg. https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/ec2_instance.rb#L235-L237) and make source an absolute reference to the file?

lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -0,0 +1,29 @@
...
+    
+  mk_resource_methods
+
+
+  def exists?
+    Puppet.info("Checking if docker-compose.yml exists")
+    begin
+      Dir.chdir(resource[:source])
+      File.file?('docker-compose.*')    
+    rescue Exception => e
```
garethr: the exception is stored in `e` here but is then not used. It's probably best to do this without an exception as well, as not finding a file isn't _exceptional_.

lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -0,0 +1,29 @@
...
+  def create
+     Puppet.info("running docker-compose up -d")
+     Dir.chdir(resource[:source])
+     dockercompose('up', '-d')
+  end
+   
+  def destroy
+    Puppet.info("stoping docker-compose containers")
+    Dir.chdir(resource[:source])
+    dockercompose('stop')
```
garethr: stop will stop the containers from running but not remove them. I'm not sure the semantics are quite right here. I think if you are only supporting absent and present for ensure to begin with then absent should remove the containers completely.

lib/puppet/type/docker_compose.rb
```diff
@@ -0,0 +1,18 @@
+Puppet::Type.newtype(:docker_compose) do
+    @doc = "Runs docker-compose"
+
+    ensurable
+
+    newparam(:name) do
```
garethr: Given this isn't used for anything else in the provider I would probably make this the file path to the compose.yml file, which removes the need for the additional source property.

manifests/init.pp
```diff
@@ -143,6 +143,10 @@
 #   Specifc a array of users to add to the docker group
 #   Default is empty
 #
+# [*docker_compose*]
+#   Specifc if you want docker-compose installed
```
garethr: Tyo for `specify`

manifests/init.pp
```diff
@@ -143,6 +143,10 @@
 #   Specifc a array of users to add to the docker group
 #   Default is empty
 #
+# [*docker_compose*]
```
garethr: I'd be tempted to rename this to manage_docker_compose to mirror other params.

manifests/install.pp
```diff
@@ -118,6 +118,72 @@
           ensure => $docker::ensure,
           name   => $dockerpackage,
         }
+      }
+    }
+      
+    if $docker::docker_compose {
```
garethr: Mmm, so I would be tempted to pull this out into a separate class, or even a separate module to begin with.

* Having the docker package explicitly manage the python-dev tools isn't a good idea, it prevents them being managed elsewhere - for instance in the [python module](https://forge.puppetlabs.com/stankevich/python).
* Ditto the pip dependencies
* EPEL is not optional based on a passed param

I don't think for the moment this should be included. I'm happy for installing docker-compose to be an exercise left up to the user, maybe with an example of how to do that in puppet I think.

lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -0,0 +1,29 @@
+Puppet::Type.type(:docker_compose).provide(:ruby) do
+  desc "Support for Docker Compose"
+  commands :dockercompose => "docker-compose"
+    
+  mk_resource_methods
+
+
+  def exists?
```
garethr: Thinking more about this, should this represent the presence of a file as it does now, or represent the fact the stack of containers it represents are running?

lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -0,0 +1,29 @@
...
+      File.file?('docker-compose.*')    
+    rescue Exception => e
+       false 
+    end
+  end
+ 
+  def create
+     Puppet.info("running docker-compose up -d")
+     Dir.chdir(resource[:source])
+     dockercompose('up', '-d')
```
garethr: If run this would stop and recreate the containers every time Puppet runs I think? Passing `--no-recreate` is likely advisable.

lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -0,0 +1,29 @@
+Puppet::Type.type(:docker_compose).provide(:ruby) do
+  desc "Support for Docker Compose"
+  commands :dockercompose => "docker-compose"
```
garethr: The module already supports passing several environment variables (like DOCKER_HOST) in via params. Ideally we'd need to support these (or at least the most common ones) for the compose CLI too.

garethr:
Hi @scotty-c. Thanks for this. I've left lots of comments, it looks a good starting point.

One think I think it should definitely support is the scale options for compose. So you could specify the number of instances of each container in Puppet (probably in a hash) and have compose scale up and down.

Any questions about the comments I've left just let me know. And thanks again.

scotty-c:
@garethr I have updated as per your suggestions above.

salimane:
I can wait for this to be merged :+1:

scotty-c:
Re opening new pull request

radiumx3:
Any ETA '?

PR #255, opened 2015-05-26
Fix a typo in selinux-enabled parameter for a RHEL/Centos 7

pdaukintis:


garethr:
@pdaukintis good catch, thanks for spotting that.

PR #256, opened 2015-05-27
Quote credentials in 'docker login' exec

jsok:
If your password (or any other field) contains characters which require escaping the `docker login` exec will fail.

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -38,7 +38,7 @@ define docker::registry(
 
   if $ensure == 'present' {
     if $username != undef and $password != undef and $email != undef {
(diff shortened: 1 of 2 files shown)


manifests/registry.pp
```diff
@@ -38,7 +38,7 @@
 
   if $ensure == 'present' {
     if $username != undef and $password != undef and $email != undef {
-      $auth_cmd = "${docker_command} login -u ${username} -p ${password} -e ${email} ${server}"
+      $auth_cmd = "${docker_command} login -u '${username}' -p '${password}' -e '${email}' ${server}"
```
pkill: Special characters didn't occur to me, thanks for the fix!

garethr:
@jsok thanks, much appreciated.

PR #258, opened 2015-05-29
make repo_opt customizable

coreone:
The "repo_opt" option that is automatically determined in params.pp works on normal RHEL7, but not on RHEL7 that is joined and controlled by a Satellite installation.  The repository names there are different than the stock RHEL7 ones.  Therefore, this pull request will keep the default normally, but allows someone to override by setting it manually.

Plus minor spelling correction...


garethr:
@coreone much appreciated. Thanks.

PR #259, opened 2015-05-29
Exports are considered invalid in systemd

cwood:
If we try to export the http/https proxys systemd will ignore them


garethr:
Is this also true for the non-rhel7 config file as well? https://github.com/garethr/garethr-docker/blob/master/templates/etc/sysconfig/docker.erb 

And what about the `no_proxy` option just below your change?

Thanks.

garethr:
Should be resolved in https://github.com/garethr/garethr-docker/pull/264. Thanks for reporting.

PR #260, opened 2015-06-04
changed docker::{image, registry, run} to use docker::command and doc…

knifhen:
…ker::service_name instead of docker::params::command and docker::params::service_name.

This is done in order to make it possible to override the command and service_name and make it possible to install and run lxc-docker on debian instead of docker.io


garethr:
@knifhen mmm, thanks for this. I noticed that the unit tests currently fail which suggests something is amiss. If you could take a look? Thanks.

PR #261, opened 2015-06-04
Added OL6 compatibility.

mbert:
Here's the patch for OL6 compatibility. I've tested it in our environment.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -72,7 +72,11 @@ class docker::params {
       $docker_command = $docker_command_default
       if versioncmp($::operatingsystemrelease, '7.0') < 0 {
         $detach_service_in_init = true
-        $docker_group = $docker_group_default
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -90,6 +94,9 @@
         } else {
           $repo_opt = undef
         }
+      } elsif (versioncmp($::operatingsystemmajrelease, '6') == 0 and $::operatingsystem == 'OracleLinux') {
```
DavidS: It might be better to write this test similar to the ones in L63 and L73:

```
if versioncmp($::operatingsystemrelease, '7.0') < 0 ...
```

assuming, there is no OL5. Or at least nobody is trying to run docker on it.

DavidS:
:+1: from me.

garethr:
@mbert thanks for adding.

PR #264, opened 2015-06-08
Remove exports from rhel7 sysconfig

### LINKED ISSUE #244: systemd new proxy configuration in RH7/Fedora
Setting a proxy in /etc/sysconfig/docker is no longer read by docker. Instead it uses the new systemd configuration detailed here: http://docs.docker.com/articles/systemd/#http-proxy

Offline testing requires the /etc/systemd/system/docker.service.d directory to be created, with the http-proxy.conf file populated as per the documentation.

This also seems to affect NO_PROXY.

I have noticed this both on RH7 and Fedora boxes.

michaelwells:
systemd does not execute files provided by the EnvironmentFile option in
the unit file and instead expects them to be a simple set of variable
assignments. This means that export commands are ignored. Replacing them
with variable assignment fixes the issue.

This should resolve #244 and covers #259 as well

### DIFF
--- a/spec/classes/docker_spec.rb
+++ b/spec/classes/docker_spec.rb
@@ -364,6 +364,11 @@ describe 'docker', :type => :class do
     it { should_not contain_class('epel') }
     it { should contain_package('docker').with_install_options('--enablerepo=rhel7-extras') }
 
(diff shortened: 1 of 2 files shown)


garethr:
Thanks @michaelwells, in particular for the explanation.

PR #265, opened 2015-06-09
Fix docker service template

bryanjbelanger:
I don't believe the service User should be changed when you try to run the container as a certain user.

For example I want the jenkins container to run as the jenkins user. However when I pass username to run.pp this template also attempts to change the "Service" user to jenkins:

User=jenkins

I nuked that like in the template.


garethr:
@bryanjbelanger mmm, I think I agree. At the very least this is something that shouldn't be coupled. Thanks

PR #266, opened 2015-06-10
Clean up older docker images left behind

cwood:
On pulling new images we will need to clean up the older ones. This is a PR that makes a cron that runs every day at 1am and removes docker images safely.

### DIFF
--- a/manifests/cleanup.pp
+++ b/manifests/cleanup.pp
@@ -1,0 +1,20 @@
+#
+#   Colin Wood <cwood@atlassian.com>
+#   10-06-15
+#
+#   Clean up old docker images safely that have been removed.
+#
+class docker::cleanup {
+  cron { 'docker rm':
+    user    => 'root',
+    command => 'docker rm $(docker ps --filter status=exited -q 2>/dev/null) 2>/dev/null',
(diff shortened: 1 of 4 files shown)


dvorak:
+1

manifests/cleanup.pp
```diff
@@ -0,0 +1,20 @@
+#
+#   Colin Wood <cwood@atlassian.com>
+#   10-06-15
+#
+#   Clean up old docker images safely that have been removed.
+#
+class docker::cleanup {
+  cron { 'docker rm':
+    user    => 'root',
+    command => 'docker rm $(docker ps --filter status=exited -q 2>/dev/null) 2>/dev/null',
```
garethr: the command to run isn't always docker

garethr:
Mmm, I'm not 100% sure on adding this to the module. Although it's obviously completely valid I can think of other strategies for this and baking one into the module when it might not be the default for everyone feels wrong. I'll leave open to ponder on. This might make a better addition as an example? ie. create an examples folder and demonstrate this?

cwood:
I think Im just gunna close it thought it would be usefull but ill just move this

PR #268, opened 2015-06-15
Exported home path in the init.d template file

rasjoh:
Fix for using the module with Puppet on a Centos machine
The /sbin/service doesn't retain the home variable


garethr:
@rasjoh thanks, appreciate the fix

PR #269, opened 2015-06-15
Switch to using >= 2.0.0 for Apt module

protochron:
This updates Apt::Source to be compatible with the new version of the Puppet Labs Apt module. I hope this doesn't break stop anyone else from upgrading this module!


protochron:
Hm, this doesn't seem to work with the future parser or Puppet 4.0 although the upstream apt module does. I need to spend more time figuring out why.

PR #271, opened 2015-06-19
Execute command before stop

jovandeginste:
Add (optional) command to execute before shutting down a container:

docker::run { "myinstance":
	...
	before_stop => "/path/to/my_clean_shutdown_commmand",
}


jovandeginste:
@garethr anything wrong with this PR?

PR #274, opened 2015-06-20
upgrade apt module

jhoblitt:


garethr:
@jhoblitt thanks, I'd been meaning to update that now that 2.1 is backwards compatible. Much appreciated.

jhoblitt:
@garethr Thanks!  Somewhat ironically, it was preventing me from testing an env on Fedora 22.  We really need to figure out an optional dependency mechanism...

PR #280, opened 2015-06-25
(MAINT) Adds beaker/puppet_install_helper gem

bmjen:
This will allow the module to be easily tested by Puppet jenkins ci,
and facilitates testing with foss, pe, and agent environments.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,17 +1,9 @@
 require 'beaker-rspec/spec_helper'
 require 'beaker-rspec/helpers/serverspec'
+require 'beaker/puppet_install_helper'
 require 'pry'
 
-unless ENV['BEAKER_provision'] == 'no'
-  hosts.each do |host|
-    # Install Puppet
-    if host.is_pe?
-      install_pe
-    else
-      install_puppet
-    end
-  end
-end
+run_puppet_install_helper
 
 RSpec.configure do |c|
   # Project root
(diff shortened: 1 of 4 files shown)


mhaskel:
:+1:

spec/spec_helper_acceptance.rb
```diff
@@ -1,17 +1,9 @@
...
-  hosts.each do |host|
-    # Install Puppet
-    if host.is_pe?
-      install_pe
-    else
-      install_puppet
-    end
-  end
-end
+run_puppet_install_helper
```
garethr: What's the default behaviour for this? Just checking this maintains backwards compatibility at least in terms of a) installing open source puppet and b) working on the vagrant hypervisor

bmjen: The puppet_install_helper gem will default the install type to 'foss'.
https://github.com/puppetlabs/beaker-puppet_install_helper/blob/master/lib/beaker/puppet_install_helper.rb#L82

And it works with vagrant.

garethr:
Note that this should have had an update to the docs to indicate you now need to pass `PUPPET_VERSION` to every run. I'd prefer that to have a default.

garethr:
@bmjen ah, worse than that. At least for me locally this doesn't run, at least with the latest version of beaker. Specifically it fails like:

    Users/garethr/.rvm/gems/ruby-2.1.4/bundler/gems/beaker-a14f162484a9/lib/beaker/host.rb:273:in `exec': Host 'ubuntu-server-1404-x64' exited with 8 running: (Beaker::Host::CommandFailure)
 wget -O /tmp/puppet.deb http://apt.puppetlabs.com/puppetlabs-release-.deb

PR #282, opened 2015-06-26
Change installation URL from get.docker.io to get.docker.com

HackAttack:
On Ubuntu I have to use get.docker.com/ubuntu in the apt source to get the latest version. Using docker.io tells me that 1.5 is latest instead of 1.7 as expected.


garethr:
@HackAttack thanks, much appreciated.

PR #283, opened 2015-06-27
Rename docker::run to docker::container to prepare for ensure parameter

salimane:
As per the comment https://github.com/garethr/garethr-docker/issues/130#issuecomment-60485838, in order to add the `ensure` parameter, this was needed.
This pr includes:
- Rename `docker::run` to `docker::container`
- Make sure it's still backwards compatible for those still using `docker::run`
- Added a test to make sure the backward compatibility is working.

@garethr what do you think ?

Thanks


garethr:
@salimane thanks for looking at this. I'm travelling a bit at the moment so won't merge immediately, but I would like to make this switch probably as part of a 4.0 release of the module in the not-too-distant future.

PR #284, opened 2015-06-28
Support installing from RPM url on RedHat based systems. Fixes #277

### LINKED ISSUE #277: Support installing from RPM url on RedHat based systems
This would be useful for getting the latest version of Docker provided by the project.

salimane:
On Centos 6.x at least, the docker upstream version is maxed out at 1.5 . This pr allows to install custom rpm packages from docker.io (1.6, 1.7) on Centos based systems. It fixes #277 .
ping @garethr

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,18 @@ class { 'docker':
 }
 ```
 
+And if you want to install a specific rpm package of docker you can do so:
+
+```puppet
+class { 'docker' :
+  manage_package              => true,
+  use_upstream_package_source => false,
+  package_name                => 'docker-engine'
+  package_source              => 'https://get.docker.com/rpm/1.7.0/centos-6/RPMS/x86_64/docker-engine-1.7.0-1.el6.x86_64.rpm',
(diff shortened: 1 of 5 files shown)


garethr:
@salimane thanks for submitting this change. Apologies for taking quite so long to getting to reviewing it.

I notice that this change breaks the behaviour around `$docker::repo_opt`, specifically the behaviour in master is to only pass the install options when this is set. The diff here means that for both package declarations it's always passed? https://github.com/salimane/garethr-docker/blob/master/manifests/install.pp#L112-L118. This might just be case of rebasing the change against master?

If you could reintroduce that logic that would be great.

Also, it would be useful to add a quick test to the unit tests to cover this case, and the resulting logic. The unit tests really help with maintaining existing functionality from regressions. The CONTRIBUTING.md doc has the basic details but if you have any questions or need a hand do let me know. And thanks again.

salimane:
@garethr I've rebased the PR. I've also reintroduce the ``$docker::repo_opt`` logic. A test was also added to check this.
Thanks

garethr:
@salimane thanks, much appreciated. Nice addition.

PR #285, opened 2015-06-29
extra_parameter do not work when docker::run restart policy set

jsok:
The `extra_parameters` parameter has no effect with a restart policy is set. e.g.

```puppet
  docker::run { 'sample':
    image            => $image,
    extra_parameters => ['--log-driver=syslog'],
    restart          => 'no',
  }
```

In this example, the `--log-driver` option will never be passed to `docker run ...`.

This PR simplifies the existing `extra_parameters` handling by moving it into `docker_run_flags`. This way the `docker-run` init templates can be simplified and also fixes the above issue.


garethr:
@jsok thanks, much appreciated. Good spot.

Issue #286, opened 2015-06-29
docker::run { restart => anything } should use title for container name


jstangroome:
If any value is specified for the `restart` parameter of the `docker::run` class then the sanitised title is used for the CID file but it should also be used as the container name, potentially based on whether the `use_name` parameter has also been set.

garethr:
I think this is now fixed. If you still see the problem let me know and I'll delve in. `use_name` is no-longer required which simplifies some of the usage here.

PR #287, opened 2015-06-29
Set the container name if the `restart` parameter is specified.

jstangroome:
And also the `use_name` parameter is true. Refs #286.


jstangroome:
@garethr I've merged latest, resolved the conflict and added a test for the new logic. Please review.

garethr:
@jstangroome apologies for taking a while to get back to you on this issue. As of the latest release `use_name` is no longer required, in fact it issues a deprecation notice. So this issue should be resolved all being well. Thanks for submitting anyhow.

PR #290, opened 2015-07-02
Added Docker-compose

scotty-c:
Good morning Gareth,

Here is the re write for docker-compose functionality. I have added a few more features and cleaned up the provider


radiumx3:
ETA ?

scotty-c:
Moving to Docker-compose branch

PR #291, opened 2015-07-02
Fix when running=false : do not start the docker image on docker restart

ziporah:
When performing a systemctl restart docker, all images even the ones
with running=false are being started. This is unexpected behaviour.
The systemctl restart is needed when using docker in combination with
firewalld. After every firewalld restart, docker needs a restart in
order to make sure the containers keep working. The problem is fixed by
disabling the specific container in systemd. This makes sure the container isn't
respawned at a systemctl docker restart.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -178,28 +178,40 @@ define docker::run(
       mode    => $mode,
     }
 
-    # Transition help from moving from CID based container detection to
-    # Name-based container detection. See #222 for context.
-    # This code should be considered temporary until most people have
-    # transitioned. - 2015-04-15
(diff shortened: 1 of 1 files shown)


garethr:
@ziporah thanks for spotting this. A related PR just made this more configurable, but unfortunately this PR no-longer merges cleanly into master. If you could rebase I'll get this merged in.

Also note that running could also be stopped as well as false, so probably best to check for that as well.
https://docs.puppetlabs.com/references/latest/type.html#service-attribute-ensure

ziporah:
@garethr I fixed the conflicts and the code can be merged now with master.
The running = stopped or running = running  value can not be set with the current code base, this needs refactoring of the validation as well. If all works fine, I will add this later in another PR

PR #292, opened 2015-07-03
Improve docker::run service configuration

dvorak:
Added two new parameters:

The `service_prefix` parameter allows overriding the prefix that is
applied to the service resource and the init script.  This makes it
easier to integrate with existing modules that have specific service
names in place.

The second parameter 'manage_service' allows turning off the creation of
the service resource.  Again, this makes it easier to integate with
existing modules that already manage the resource.  The restart_service
parameter will continue to notify the service if it is found under a
matching name.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -18,6 +18,19 @@
 # This will allow the docker container to be restarted if it dies, without
 # puppet help.
 #
+# [*service_prefix*]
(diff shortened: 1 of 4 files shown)


garethr:
@dvorak thanks, a nice addition. And thanks for adding the tests and docs too. Much appreciated,

PR #296, opened 2015-07-09
Update README.md

activars:
Correct documentation for run_instance hiera data

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,10 +242,14 @@ docker::run { 'helloworld':
 If using hiera, there's a docker::run_instance class you can configure, for example:
 
 ```yaml
-docker::run_instance:
-  helloworld:
-    image: 'ubuntu:precise'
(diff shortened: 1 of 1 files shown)


garethr:
I think @ricoli originally added this, as I'm not using this directly at the moment could you confirm if this is correct?

ricoli:
Indeed I did add this section. Didn't add the class include using hiera as not everyone likes to do so, but I guess it doesn't hurt to have it in the readme for clarity sake :)

ricoli:
Oh and I just realised that it should actually be a hash of instances, so the way I wrote it is correct. For clarity, the class should have $instances = {}

ricoli:
Also, the class and readme should also be updated to have $instances = {}, for clarity sake. My bad there!

activars:
The variable is named $instance ...

ricoli:
Indeed @activars, and it was my mistake to do so - should have been instances, as it's a hash :)

PR #297, opened 2015-07-13
Add wrapper for docker run commands

claytononeill:
This adds a new defined type that follows the pattern that the
docker::run defined type has already set out.  Instead of generating an
init script, it generates a shell wrapper to run a command via docker
run.  By default, the container it is run in will be unnamed and deleted
after the command exists.  This can is useful when converting existing
tools from packages to docker containers since it can allow them to
continue running the same commands via the generated wrappers.

### DIFF
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -48,6 +48,14 @@ module Puppet::Parser::Functions
       flags << '-t'
     end
 
+    if opts['rm']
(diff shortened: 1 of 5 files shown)


dvorak:
Updated against latest master and now passes CI for the PR's 1 year anniversary!  @garethr can you take a look at this?

claytono:
Going to close this.  We're no longer using this module and it seems Gareth is busy with other things.  Someone else can resubmit if they need it.

PR #298, opened 2015-07-14
Install net-tools package where needed in tests

DavidS:


DavidS:
This is the same problem as https://github.com/puppetlabs/puppetlabs-postgresql/pull/651

PR #299, opened 2015-07-15
Supporting Ubuntu 15.04 and removing direct apt include

narkisr:
This pull request enables systemd for Ubuntu 15.04 and removes the direct include of apt (thus preventing collisions with other modules)

Thanks


garethr:
Thanks for looking at this. I'm afraid the unit tests fail due to the removal of apt.

The use of include should mean this doesn't clash with other modules? That's the canonical usage of include https://docs.puppetlabs.com/puppet/latest/reference/lang_classes.html#using-include

If you're having an issue with that if you could provide a simple test case?

If you want to submit the support 15.04 support in a sepaate PR that might be advisable? If you could update the unit tests as well to verify that too. The [Contributing](https://github.com/garethr/garethr-docker/blob/master/CONTRIBUTING.md) guide has some info but feel free to ask any questions too.

narkisr:
The main issue I had with include apt was that using an apt class with custom parameters in other modules caused a collision with the include.

Any case ill separate the 15.04 from the apt include removal

narkisr:
Any idea on how I can remove the commit without creating a new pull request?

garethr:
`git rebase` is your friend here. Spefically do an interactive rebase and then force push the resulting changes to your branch. Something like

```
git rebase -i HEAD~2
<delete the line for the commit you want to remove>
git push origin <branch-name> --force
```

narkisr:
Ok, just did (usually rebasing published commits isn't recommended but on this case it makes sense)

Thanks

narkisr:
Any reason why this isn't merged? it makes the module non functional on Ubuntu 15.04

Thanks

garethr:
This should now work in master, which now has proper support for the systemd debian/ubuntu changes and the new packages. Thanks for reporting.

PR #300, opened 2015-07-15
Change docker::registry class to be idempotent

claytononeill:
This reworks the existing docker::registry class to use augeas instead
of the docker login/logout commands.  This resolves an issue with the
existing defined type where the Exec resource would fire on every puppet
run.  The docker command has no current way to determine if you've
already logged into a remote registry, but it stores those credentials
in the ~/.docker/config.json file.  In order to ensure the
~/.docker/config.json file and directory exist, they will be created
with exec resources, but only if they are missing.  This allows
bootstrapping new environments without manual intervention.

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -1,4 +1,4 @@
-# == Class: docker
+# == Class: docker::registry
 #
 # Module to configure private docker registries from which to pull Docker images
 # If the registry does not require authentication, this module is not required.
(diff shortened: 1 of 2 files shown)


garethr:
While augeus is in the new puppet-agent packages it's an optional install before then, and ensuring that it's available isn't handled here. I've always seen augeas as a bit love it/hate it - so I'm currently unsure about requiring it for this module. I'll have a think, and open to comments here.

dvorak:
The only other way I can think of to solve this idempotently is to use a template, and that's pretty gross, and could potentially damage any non-managed config.son data.  I don't know of any now, but it's a possibility.

dvorak:
@garethr Any more thoughts on this?  It's been outstanding for almost a year now, and I think there is real value in having an idempotent way of handling docker logins.

telegrapher:
Since it seems @garethr was undecided, I'll give my POV.

I work with a private registry, and having it without a password is a big no-no. I use (like a lot of other people) the registry with a reverse proxy in front for SSL+Basic auth.

The only reason why I don't use this module and keep my home-cooked one full of templates is actually that the docker login is not idempotent. 

I keep track of when/why Puppet does changes, and investigate when an "unexpected change" happens. This module would make that way more difficult.

Regarding augeas, it seems a frequent dependency. Probably most of the users have it there already.

walkamongus:
+1 for review

I just ran into this issue as well, although I also wish there was another way besides augeas to solve it.

dvorak:
@walkamongus The goal with using Augeas was to try to not destroy any new (possibly undocumented) config keys that docker might put in this file in the future.  It would certainly be possible to move to using some sort of template if @garethr isn't concerned about that scenario.

jaxxstorm:
Short of another solution, do we think this can be merged? the current behaviour is really annoying :)

claytononeill:
Updated to remove merge conflicts.

claytononeill:
The travis failure is unrelated to my change, and due to the fact the PuppetSyntax check recently changed to be fatal on warnings.  Since there is a Ruby 1.9 deprecation warning, it will always fail on 1.9 until that is addressed.

claytononeill:
@garethr Are there further changes you'd like to see to this?  It seems like a lot of people want something that meets this criteria, even if this implementation might need to be tweaked.

antaflos:
Requiring Augeas is really a trivial price to pay for finally getting this module to work in an idempotent way, especially since idempotency is one of the cornerstones of Puppet itself.

Other, more intrusive and less flexible ways to manage the contents of `~/.docker/config.json` would require something like [Concat](https://forge.puppet.com/puppetlabs/concat) or [Datacat](https://forge.puppet.com/richardc/datacat). Augeas is preferable because it allows to just manage the `auths` section and leaves the user free to manage other aspects of the configuration file manually.

Another way, without Augeas, would be to add an `unless`/`onlyif` clause to the [Exec resource in `docker::registry`](https://github.com/garethr/garethr-docker/blob/master/manifests/registry.pp#L63-L70) that does some `grep` magic on `~/.docker/config.json` to see if the authentication credentials have already been added to the file. This probably requires GNU grep or a grep variant with similar capabilities (like PCRE) to be available. I'll see if I can put together a PR that implements this so we can compare alternatives.

claytononeill:
The original complaint was while augeas is in the puppet-agent packages, it isn't universally available everywhere else.  Now that Puppet 3 is EOL, this seems to no longer be a concern.  I imagine there is someone, somewhere that's using Puppet 4 without the puppet-agent packages but it seems like requiring someone like that to also support augeas if they want to use this login type is a pretty minor requirement.

antaflos:
I agree. FWIW I think this should be merged.

I've had moderate success with an `unless` clause that uses `grep`:

```puppet
$auth_string = base64('encode', "${username}:${password}", 'strict')

...

exec { "${title} auth":
  ...
  unless => "/bin/grep -qPaz '(?s).*\"auths\": {\\n.*\"${server}\": {\\n\\s+\"auth\": \"${auth_string}\"\\n\\s+\"email\": \"${email}\"' ~${local_user}/.docker/config.json"
  ...
}
```

This succeeds in keeping the Exec resource from refreshing but I think Augeas is a much cleaner solution to this and doesn't require GNU grep and complex regular expressions.

One thing I noticed is that the `email` flag/attribute is deprecated and will be removed in Docker 1.13:

> Flag --email has been deprecated, will be removed in 1.13.

You could simplify the Augeas call and just ignore the `email` parameter.

olivierHa:
Hello, Any updates on this subject? This seems stuck for almost 2 years :(

claytono:
We're no longer using this module and it seems Gareth is busy with other things.  Someone else can resubmit if they need it.

telegrapher:
Probably a better place to submit anything will be: https://forge.puppet.com/puppetlabs/docker/readme

PR #301, opened 2015-07-17
Fix memory_limit format on README.md

orifito:
`memory_limit` values must be specified within single quotes


garethr:
@orifito thanks for spotting, much appreciated.

orifito:
@garethr thanks to you for such a great puppet module! :smiley:

PR #302, opened 2015-07-17
fix Debian support 

deric:
* command is called `docker`
 * package is called `lxc-docker`
 * allow changing parameters on any Debian based distro

Current implementation doesn't allow to modify `$package_name` simply by passing different `$package_name_default` on Debian.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -38,18 +38,9 @@ class docker::params {
   $docker_group_default         = 'docker'
   case $::osfamily {
     'Debian' : {
-      case $::operatingsystem {
-        'Ubuntu' : {
-          $package_name   = $package_name_default
-          $service_name   = $service_name_default
-          $docker_command = $docker_command_default
-        }
-        default: {
-          $package_name   = 'docker.io'
-          $service_name   = 'docker.io'
-          $docker_command = 'docker.io'
-        }
-      }
+      $package_name   = $package_name_default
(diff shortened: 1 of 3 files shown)


garethr:
@deric thanks for this. What version of Debian are you testing? And are these the docker inc packages or are these in the main Debian repos?

Note also that `package_name` can be used in the main docker class to override this. Params is basically private as implemented and intended for module level defaults.

Note as well that Docker now have official debian repos so I'm currently in the process of adding proper support for these. At which point I'm hoping to add Debian to the supported list.

deric:
@garethr packages are from docker.io, I've tested this on Wheezy and Jessie (Debian 8). AFAIK there are no Docker packages in Debian upstream.

garethr:
Ah, docker.io is actually deprecated (https://github.com/docker/docker/pull/14920). The good news is new repos are available with full Debian support (http://blog.docker.com/2015/07/new-apt-and-yum-repos). I'm currently working on https://github.com/garethr/garethr-docker/pull/311 which should enable these, and I'll test this out on Debian. Assuming I can get this working I'll add Debian to the acceptance tests too.

garethr:
Thanks for the work on this. I've got Debian supported (both 7.8 and 8.0) on the PR in #311 so I'm going to close this PR. Do try out the work in that PR, which should be merged shortly

PR #304, opened 2015-07-20
add log-driver run option

ryfow:


README.md
```diff
@@ -228,6 +229,9 @@ Specifying `before_stop` will execute a command before stopping the container.
 
 The `depends` option allows expressing containers that must be started before. This affects the generation of the init.d/systemd script.
 
+The `log_driver` option is passed as the --log-driver option to Docker. As of Docker 1.6, `json-file` and `syslog` are
```
garethr: It would be useful to add these as validation rules to the type. What you describe is true of Docker, rather than of the module at present.

garethr:
Hi @ryfow, thanks for adding this. It would be useful to add some basic unit tests (using rspec-puppet) to this PR. These help ensure functionality like this isn't broken in a future release, especially with the large number of contributions. If you need a hand with those let me know, you can find some details in the CONTRIBUTING file too.

garethr:
@ryfow thanks for submitting the PR. Just to note I've merged https://github.com/garethr/garethr-docker/commit/9ed84678aa2e0058a1bbbc86a5b4786adb6997b9 which contains a pretty complete log-driver implementation.

PR #306, opened 2015-07-22
Add acceptance tests for repository type

garethr:


spec/acceptance/docker_spec.rb
```diff
@@ -95,6 +97,51 @@ class { 'docker':
...
+    it 'should be able to login to the registry' do
+      manifest = <<-EOS
+        docker::registry { '#{@registry_address}':
+          username => 'user',
+          password => 'password',
+          email    => '#{@registry_email}',
+        }
+      EOS
+      apply_manifest(manifest, :catch_failures=>true)
+      shell("cat ~/.docker/config.json | grep #{@registry_address}", :acceptable_exit_codes => [0])
```
DavidS: `grep PATTERN FILENAME` is usually preferable to `cat | grep`, also quoting the pattern would be good practice, although I can see it's not required here.

spec/acceptance/docker_spec.rb
```diff
@@ -1,12 +1,12 @@
 require 'spec_helper_acceptance'
 
-describe 'docker class' do
+describe 'docker' do
   case fact('osfamily')
   when 'RedHat'
-    if fact('operatingsystemrelease').to_f >= 7
-      package_name = 'docker'
+    package_name = if fact('operatingsystemrelease').to_f >= 7
```
DavidS: I don't like if-stmts as values. I'd prefer either the original version or using a ternary. I'd defer to "local style", tho.

spec/acceptance/docker_spec.rb
DavidS: I'd expect acceptable_exit_codes default to default to `[0]`, in which case you can leave it off.

spec/acceptance/docker_spec.rb
```diff
@@ -95,6 +97,51 @@ class { 'docker':
...
+      shell("cat ~/.docker/config.json | grep #{@registry_email}", :acceptable_exit_codes => [0])
+    end
+
+    it 'should be able to logout from the registry' do
+      manifest = <<-EOS
+        docker::registry { '#{@registry_address}':
+          ensure=> absent,
+        }
+      EOS
+      apply_manifest(manifest, :catch_failures=>true)
```
DavidS: No idempotency checking here or above.

spec/acceptance/docker_spec.rb
garethr: I _really_ dislike the fact their is no explicit assertion here, ie. that shell has a side effect. The best I can do (without accepting all exit codes) is at least make it explicit what will cause shell to fail. Without that, as a reader, I would likely miss the fact that shell can fail the test for reasons other than an actual error.

PR #311, opened 2015-07-26
Support new Docker repos

garethr:
Docker have created a new set of official repositories, including adding
Yum and Debian based packages.


rpete:
ping :+1:

stevenschlansker:
Would be nice to see a release with this.
Also please consider #322 which is a PR to this PR.

stevenschlansker:
I see that this is still in `manifests/install.pp`:

```
 if $docker::version {
  $dockerpackage = "${docker::package_name}-${docker::version}"
} else {
  $dockerpackage = $docker::package_name
}
```
Shouldn't this be gone, now that there is just the `docker-engine` package with proper versioning?

JamieCressey:
Any idea when this will get merged? I'm using a local hacked version currently to support the new repos, but would be nice to revert to the upstream copy of the module.

bydga:
+1

stevenschlansker:
tag @garethr any news on this moving forward?  Thanks!

garethr:
Still working on this, ensuring everything still works across the different supported platforms is taking a little time unfortunately and I only have a limited amount of time outside work and being on holiday for the last week or so. I'm hoping this will land in the next week or two all being well.

garethr:
So, consider the existing PR a release candidate. If anyone has a moment to test this out that would be much appreciated. I'll hopefully merge and release next week.

azweb76:
Tested on CentOS 7. Initial run fails to apply dropin before start of the service. If I run `systemctl restart docker`, I see proper args appended to process in `ps aux | grep docker`. Otherwise, works great.

stevenschlansker:
Smoke test passes here

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.2 LTS"
```

heterodox:
Tested on CentOS 7; with SELinux enforced and Docker 1.8.x(?) the docker.service start will time out unless the docker-selinux package is installed. We had already noticed this when using the new repos. When we added the following to our node manifest:

	package { 'docker-selinux':
		ensure => 'present',
		before => Package['docker']
	}

The service then changed to 'running' successfully.

ngiger:
I tried you branch with the following example:

> include docker
> docker::image { 'debian': }
> docker::run { 'helloworld':
>   image   => 'debian',
>   command => '/bin/sh -c "while true; do echo hello world; sleep 1; done"',
> }

and got the following error `Error: /Stage[main]/Main/Docker::Run[helloworld]/Service[docker-helloworld]: Could not evaluate: Could not find init script for 'docker-helloworld'`

The following patch fixed this:

```
diff --git a/manifests/run.pp b/manifests/run.pp
index cbbab48..1d75dfd 100644
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -220,10 +220,16 @@ define docker::run(
         } ->
         File[$initscript]
       }
+      if $uses_systemd {
+        $provider = 'systemd'
+      } else {
+        $provider = 'default'
+      }
 
       service { "${service_prefix}${sanitised_title}":
         ensure    => $running,
         enable    => true,
+        provider  => $provider,
         hasstatus => $hasstatus,
         require   => File[$initscript],
       }
```

stevenschlansker:
If you attempt to specify a version manually, it looks like the package name ends up broken:
```
E: Unable to locate package docker-engine-1.8.2-0~trusty
```

zahaim:
Tested on Ubuntu 14.04.1 LTS. 
Puppet v3.8.2 with hiera
No issues so far.

hdeadman:
I am using this on Centos 7 & puppet 4 and its installing docker 1.8 from the upstream repos, it's working fine - running with this branch plus latest commits on master.

jae2:
I've run this branch on Wheezy with puppet 3.8.2 and haven't noticed any issues so far.

garethr:
Thanks @ngiger for that patch.

jae2:
One last thing - I noticed that this will only work on wheezy if you're using >= 3.16 kernel. This is in the docker pre-reqs in the wheezy installation guide so I suppose it shouldn't be an issue for this module. Without this docker::exec will fail without an error message but throw error code 255.

mjasion:
@garethr When do you plan to publish this PR on puppet forge? :-)

PR #313, opened 2015-07-27
Fact to report the number of running containers

tux-o-matic:
Simple Fact using 'docker ps' to report with Facter the number of running containers.
Since there are no Fact in this module today, hints or an update to the PR guidelines around Facter testing could be nice.

### DIFF
--- a/lib/facter/docker_running_container.rb
+++ b/lib/facter/docker_running_container.rb
@@ -1,0 +1,18 @@
+
+# Fact: docker_running_container
+#
+# Purpose: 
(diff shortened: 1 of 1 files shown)


garethr:
@tux-o-matic thanks for submitting and just to say I've not completely forgotten about this. I'm going to spend some time all-being-well on what facts would make sense to expose for Docker above and beyond just exposing the number of containers. With that in mind I'll hold of merging this so as to avoid an issue with that interface later. Thanks for the idea anyhow, much appreciated.

PR #315, opened 2015-08-03
Updating to avoid key_source and include_src depreciation warnings from puppetlabs-apt

ChrisCrewdson:
I am not sure if this change needs to be locked to a specific puppetlabs-apt version, but it passes all the current tests.


garethr:
@ChrisCrewdson thanks. This does indeed only work with the 2 series of the apt module.

I'm not going to merge this at this point as doing so forces anyone using this module to update the apt module. I'd rather wait for that to bed down a little and then make the change given it's just implementation. 2.1 of the apt module is backwards compatible so it's not a problem apart from the deprecation warnings.

ColinHebert:
It's been quite a while now. Should this be merged? (2.1 has been released a year ago)

erezny:
This pull request is out of date and no longer works on master. Use https://github.com/garethr/garethr-docker/issues/459 Instead.

ChrisCrewdson:
Closing this very old PR in favor #459.

PR #317, opened 2015-08-11
Add 'after' parallel to 'depends'

jovandeginste:
* after only implies order
* depends implies two-way dependency


jovandeginste:
We ran into a problem where we needed one container ('registrator') to be started before others, but restarting the registrator would restart all depending containers too (which was not what we wanted). Splitting 'after' and 'depends' should fix this issue.

garethr:
@jovandeginste this looks good. I want to get some time to test out but thanks for submitting and apologies for taking a while to get back to you.

jovandeginste:
Obsoleted by #393

PR #319, opened 2015-08-12
Add CentOS 7 default for repo_opt

coreone:


garethr:
@coreone Thanks for this, I think the centos box I was using already had this enabled

PR #321, opened 2015-08-13
Add labels

jovandeginste:
Add support for docker labels


garethr:
@jovandeginste ah, thanks for this. I was just starting to experiement with labels and this makes things easier. Cheers

PR #322, opened 2015-08-13
Remove apt::source ~> apt_update dependency

stevenschlansker:
The apt module already has a dependency like this, and having this
creates a dependency cycle with puppetlabs/apt 1.8.0

```
Error: Could not apply complete catalog: Found 1 dependency cycle:
(Anchor[apt::source::docker] => Apt::Source[docker] => Exec[apt_update] => Class[Apt::Update] => Anchor[apt::source::docker])
```


garethr:
Thanks for testing, I've removed this in the #311 PR.

Issue #305, opened 2015-07-21
docker::run { restart => anything } fails if container exists but is stopped


jstangroome:
If any value is specified for the restart parameter of the docker::run class then the output of `docker ps --no-trunc` is grepped for the container ID from the CID file before attempting to execute `docker run`.

However `docker ps --no-trunc` does not return stopped containers so the `docker run` command is still executed and then fails with the message:

```
Container ID file found, make sure the other container isn't running or delete <cid file>
```

As a minimum, this should not result in an error that causes the Puppet run to fail.

It would be more useful to detect stopped containers and either leave them stopped or offer to start them (perhaps based on the value of another parameter).

itsSaad:
Im still having this issue,  where i stop a running container (restart: always). puppet want to create the container again and fails because it already exists. #674

PR #325, opened 2015-08-17
Find stopped containers when restart param is given

paschdan:
this should also fix gh-305


garethr:
@paschdan thanks, much appreciated.

PR #327, opened 2015-08-18
Fixes to new docker repos

nlloyd:
In the interest of keeping current with docker, I grabbed this branch and played around with it in our puppet dev space. I found a few issues with the manifests and made the fixes here.


garethr:
Quick look at this it appears like it only supports the Ubuntu code path? `real_docker_version` won't be set anywhere else but can be accessed?

The unit tests are also broken on this branch, potentially just due to the above but as likely due to the change in behaviour.

It would be useful to mention the specific issues for the individual fixes if possible. Some of them relate to adding version support while others may be valid fixes - mixed in without context it's difficult to tell. I think the current acceptance tests run cleanly on Ubuntu 14.04?

Thanks for talking a look, and if you have chance to expand on the specific issues that would likely be very useful.

nlloyd:
Certainly, honestly I created this PR right away just to have it logged so I didn't forget! I will make the appropriate changes in the next few days.

diranged:
Hey I'm still looking forward to seeing this get finished and merged. One side note -- it would be great if we can add some kind of "don't upgrade" option. That is, if we update our puppet code to use Docker 1.8 going forward, I'd like to avoid having live machines (running some variant of 1.7) upgraded if at all possible.

nlloyd:
I completely dropped the ball on this, sorry about that. I will get cracking on it tonight and see if I can't finish it all up for this week.

nlloyd:
This is now updated with support for RedHad and Archlinux. Unit tests are also updated.

garethr:
@nlloyd thanks for updating. A few things have landed in master. Would you be able to rebase and to squash the commits? I had a quick look through but following the existing diff with https://github.com/nlloyd/garethr-docker/commit/814191032589177e17a708103532a6bacdd4615e in is proving tricky.

nlloyd:
I haven't forgotten about this (ok well maybe a little...). Working on it now-ish.

PR #331, opened 2015-08-21
Fix spec for platforms with docker 1.7

hunner:
Docker 1.7 moved the configuration from ~/.dockercfg to
~/.docker/config.json and this is now the version available on CentOS 7.
This change checks the docker version rather than the operating system
so that the check is not bound to the platform.


garethr:
@hunner thanks!

PR #334, opened 2015-08-25
fixes missing restart param to docker-run

paschdan:


garethr:
@paschdan thanks for adding, much appreciated.

PR #335, opened 2015-08-25
Added fail statements to notify user that dm_ options are ignored unl…

n0coast:
…ess storage_driver is set to 'devicemapper'


n0coast:
Had to add all the dm_ variables to service.pp for correct scoping in puppet v.4.x

garethr:
Thanks, I've merged this manually as it clashed with something else in master. Thanks again.

PR #339, opened 2015-08-31
fix overlay vs. overlayfs in storage_driver regex

robbyt:
The regex was validating for `overlay` but the documentation says that `overlayfs` is correct.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -204,7 +204,7 @@ class docker(
   }
 
   if $storage_driver {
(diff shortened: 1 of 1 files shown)


garethr:
Thanks, @robbyt . Do you have a link to the documentation in question? Just checking it's not something that's platform specific.

garethr:
@robbyt the documentation at http://docs.docker.com/engine/reference/commandline/daemon/ appears to indicate `overlay` is correct?

robbyt:
Yes, `overlay` is correct. Sorry I missed your previous message. I'll be sending a new PR with an additional test and update to the docs for overlay.

PR #340, opened 2015-09-04
Make it work for jessie

ngiger:
According to /usr/share/doc/docker.io/changelog.Debian.gz the bin/docker.io symlink was removed with docker 1.3.1

As I think few people want to install docker < 1.3.1 on debian jessie, I used jessie to reflect this change.

Now your find puppet module works fine for me under jessie!

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -45,9 +45,15 @@ class docker::params {
           $docker_command = $docker_command_default
         }
         default: {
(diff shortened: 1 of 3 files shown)


garethr:
Full support for Debian is coming in https://github.com/garethr/garethr-docker/pull/311

This features support for the new repos, and well as adds Debian to the acceptance tests.

PR #341, opened 2015-09-04
Storage options update

coreone:
There are several storage options that have been added to Docker that are not included in this module:

* dm.thinpooldev
* dm.use_deferred_removal
* dm.blkdiscard
* dm.override_udev_sync_check

Additionally, CentOS/RHEL 7 have added this `docker-storage-setup` service if you use non-loopback devices for docker storage, so I also added the `/etc/sysconfig/docker-storage-setup` config template and its associated parameters.


coreone:
Crap...bad rebase...closing and re-rebasing

PR #342, opened 2015-09-04
Docker storage settings update

coreone:
There are several storage options that have been added to Docker that are not included in this module:

* dm.thinpooldev
* dm.use_deferred_removal
* dm.blkdiscard
* dm.override_udev_sync_check

Additionally, CentOS/RHEL 7 have added this `docker-storage-setup` service if you use non-loopback devices for docker storage, so I also added the `/etc/sysconfig/docker-storage-setup` config template and its associated parameters.


coreone:
I believe this now conflicts with #346 since I had to fix a similar bug where all the config files were not being enforced before a run was scheduled, so docker would be half-configured, started, configured some more, then restart, which would break all the storage.  When #346 or a similar bug fix is added, I will revert that one change to run.pp to make this PR happy again.

manifests/run.pp
```diff
@@ -218,7 +218,10 @@
         ensure    => $running,
         enable    => true,
         hasstatus => $hasstatus,
-        require   => File[$initscript],
+        require   => [
+          File[$initscript],
+          Service[$service_name]
+        ]
```
coreone: Didn't want to edit this as part of this branch/PR, but it was the only way to get setup / docker run to work since "docker run" wasn't waiting for the service to be fully configured before attempting to run.

coreone:
Merge with upstream/master resolved conflicts, but I had to remove the require I put in `run.pp` as there was some weird dependency cycle that kept popping up in Travis.  I can't see where the cycle is coming in, but removing that require for the *docker* service causes the first run of this module to fail if a docker instance is also present in the catalog.  That first failure then makes docker unable to start until the `/var/lib/docker` directory is removed so that the metadata for Docker on that machine can be recreated.

DjxDeaf:
This is an important step since dm_datadev & dm_metadatadev are deprecated. should be added as optional.
The change is quite minor, and can be easily added - would you have time to push this? or should I suggest a pull request?

coreone:
@DjxDeaf not sure what change is being requested here, so I did a couple things.  I added some comments, added a check to make sure dm_datadev/dm_metadatadev is not used in conjunction with dm_thinpooldev, and added a notice to warn people if they were still using dm_datadev and dm_metadatadev.  Let me know if that fixes the issue you were seeing.

DjxDeaf:
@coreone You nailed it, but I think you also need to add this to the template - or did you already? as I haven't seen it in the commits.

Thanks!

templates/etc/sysconfig/docker-storage.erb
```diff
@@ -23,5 +26,9 @@ DOCKER_STORAGE_OPTIONS="<% -%>
 <% if @dm_loopmetadatasize %> --storage-opt dm.loopmetadatasize=<%= @dm_loopmetadatasize %><% end -%>
 <% if @dm_datadev %> --storage-opt dm.datadev=<%= @dm_datadev %><% end -%>
 <% if @dm_metadatadev %> --storage-opt dm.metadatadev=<%= @dm_metadatadev %><% end -%>
```
coreone: @DjxDeaf are you talking about the `dm_datadev` and `dm_metadatadev` options in the template file, because those variables already existed prior to this PR.  If not, I think I missed the problem you were stating about the template files.

DjxDeaf:
Actually i'm talking about the docker-storage file, https://github.com/broadinstitute/garethr-docker/blob/master/templates/etc/sysconfig/docker-storage.erb

dm_thinpooldev storage opt should replace the two data and metadata params.

coreone:
@DjxDeaf I didn't want to cut the other two options out entirely to prevent breaking the module for people using old versions of Docker.  This PR has the new option in `docker-storage.erb`, and the check in the manifest guaranteeing either `dm_thinpooldev` *or* the other two options should prevent the template from ever being generated with all three, or with the old options if `dm_thinpooldev` is set.

As a side note, do you know anything about Puppet 4?  I am adding tests for all these new options, and the tests work in all cases except for the 4.0.0 and 4.1.0 Puppet gems.  It seems as if Puppet 4 will not respect the if statement in the `docker-storage.erb` file:
```ruby
<% if @storage_driver == 'devicemapper' -%>
```
Has that changed in Puppet 4?  If so, is there any way to rewrite the template such that it works with both 3 and 4?

Thanks!

coreone:
Nevermind...ended up being a scoping issue where 4.x is fussier than 3.x.  I had to add all the parameters to `service.pp` since that manifest is actually the one that generates those config files from the templates.

DjxDeaf:
Yeah, scoping's a bitch.

Anyway, so what I was saying is that you just need to add an "if" for the thinpooldev so that it will exist in the config, as the data & metadata won't be.

@garethr Hey, what do you think? can we make it happen?

coreone:
@DjxDeaf were you talking about something like this in all the templates files:

```ruby
<% if @dm_thinpooldev %> --storage-opt dm.thinpooldev=<%= @dm_thinpooldev %>
<% else -%>
<% if @dm_datadev %> --storage-opt dm.datadev=<%= @dm_datadev %><% end -%>
<% if @dm_metadatadev %> --storage-opt dm.metadatadev=<%= @dm_metadatadev %><% end -%>
<% end -%>
```

DjxDeaf:
@coreone exactly. Thanks

coreone:
@DjxDeaf done

DjxDeaf:
@coreone did you send a PR?

coreone:
@DjxDeaf This is a PR...#342

DjxDeaf:
@coreone lol, i was not focused. yes, it is.. so let's wait for @garethr to take a look

coreone:
@garethr conflicts have been resolved, so this should be good to go now.

garethr:
@coreone thanks, this looks great. Much appreciated.

PR #344, opened 2015-09-15
Moved shell_values outside of OPTIONS

hanej:
Adding anything to shell_values causes it to be appended to the docker daemon options.  This fixes that by replicating what was done for docker.erb.


huddy:
Hi 

I could do with being fixed too. we're using this module and as I try and run on rhel7 I have the same problem. 

Cheers
Billy

garethr:
@hanej thanks for fixing, much appreciated.

PR #346, opened 2015-09-17
Create depency of containers on Docker service

jovandeginste:
Otherwise containers try to start before the docker service was properly configured and started...

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -31,6 +31,13 @@
 #  script.  Disabling this may be useful if integrating with existing modules.
 #  Default: true
 #
+# [*docker_service*]
+#  (optional) If (and how) the Docker service itself is managed by Puppet
+#  true          -> Service['docker']
+#  false         -> no Service dependency
+#  anything else -> Service[docker_service]
+#  Default: false
+#
 # [*extra_parameters*]
 # An array of additional command line arguments to pass to the `docker run`
 # command. Useful for adding additional new or experimental options that the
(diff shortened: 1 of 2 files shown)


jovandeginste:
I can't seem to find the cause for this error - anyone?

garethr:
Mmmm, so it would be useful to make this optional. As implemented this means you have to use the docker module `docker` class if you're using the `docker::run` define. Sometimes the former might already be available and unmanaged by Puppet.

If you could also add unit tests (some details in the CONTRIBUTING.md file) so the intent is clear and doesn't get broken in the future as the module evolves.

Thanks for suggesting. Some of the init systems deal with this already, but obviously not all. Which OS and version are you running?

zahaim:
Hello, I have this problem too (Ubuntu 14.04.1 LTS). Please let me know if you need anything more (information, tests).

jovandeginste:
@garethr I think I get your point. If it's optional, what default whould you have?

jovandeginste:
Also, I'm running this on Centos 7 64bit, with systemd.

I'm not sure this was clear, but this solves an issue only occurring during initial deployment of a machine (while using this same module for managing the docker service). So the dependencies are well set in the systemd configuration files, but the timing at initial deployment is not optimal. I can elaborate if necessary :-)

I will try to add tests later.

jovandeginste:
The functionality is now optional (defaults to false - behaves as before); the tests for each scenario should now be correct.

garethr:
@jovandeginste thanks for this , much appreciated.

PR #350, opened 2015-09-21
Adding support for unless within a docker exec type

jae2:
Currently there doesn't appear to be a way to run a docker exec command if some kind of state doesn't exist. This commit introduces support for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,10 +287,10 @@ Docker also supports running arbitrary commands within the context of a
 running container. And now so does the Puppet module.
 
 ```puppet
-docker::exec { 'bin/echo root >> /usr/lib/cron/cron.allow':
+docker::exec { 'cron_allow_root':
   detach       => true,
-  container    => 'helloworld',
-  command      => 'uptime',
+  container    => 'mycontainer',
+  command      => '/bin/echo root >> /usr/lib/cron/cron.allow',
   tty          => true,
   unless       => 'grep root /usr/lib/cron/cron.allow 2>/dev/null',
 }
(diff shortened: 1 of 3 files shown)


manifests/exec.pp
```diff
@@ -34,10 +35,12 @@
     $sanitised_container = $container
   }
   $exec = "${docker_command} exec ${docker_exec_flags} ${sanitised_container} ${command}"
+  $unless_command = "${docker_command} exec ${docker_exec_flags} ${sanitised_container} ${unless}"
 
   exec { $exec:
     environment => 'HOME=/root',
     path        => ['/bin', '/usr/bin'],
     timeout     => 0,
+    unless      => $unless_command,
```
garethr: I think this needs a guard clause? If unless isn't passed to the `docker::exec` it will be passed to the underlying exec, but with an incomplete `unless_command`.

README.md
```diff
@@ -283,15 +283,16 @@ docker::registry_auth::registries:
...
+Docker also supports running arbitrary commands within the context of a
 running container. And now so does the Puppet module.
 
 ```puppet
-docker::exec { 'helloworld-uptime':
-  detach    => true,
-  container => 'helloworld',
-  command   => 'uptime',
-  tty       => true,
+docker::exec { 'bin/echo root >> /usr/lib/cron/cron.allow':
```
garethr: This is misleading, in that this is just the title and not the command which is run on the container, which is in the `command` param.

garethr:
Great idea. Thanks for the PR. I've left a few comments, any questions let me know.

jae2:
@garethr, thanks for the feedback. I've added the changes you recommended.

garethr:
Thanks @jae2, much appreciated.

PR #351, opened 2015-09-25
Use new apt.dockerproject.org repository

### LINKED ISSUE #345: Ubuntu install does not follow Docker's guide
The install of Docker on Ubuntu 14.04 does not install from the right source. 

The source per http://docs.docker.com/installation/ubuntulinux/ is https://apt.dockerproject.org/repo and the package is docker-engine - that's what the script that gets pulled from https://get.docker.com/ does.

I was expecting this module to do the same.

gehel:
A new repository has been put in place for docker, this module should probably use it. See https://blog.docker.com/2015/07/new-apt-and-yum-repos/ for details.

Note that there is also a RPM repository, so similar changes might be needed for the RedHat part...

This fixes #345.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -51,7 +51,7 @@ class docker::params {
         }
       }
       $docker_group = $docker_group_default
-      $package_source_location     = 'https://get.docker.com/ubuntu'
+      $package_source_location     = 'http://apt.dockerproject.org/repo/'
       $use_upstream_package_source = true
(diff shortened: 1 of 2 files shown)


garethr:
Hi @gehel, you can track work on this in https://github.com/garethr/garethr-docker/pull/311. It's a bit more involved that just pointing at the new repos but that PRs is just about good to go into master.

PR #354, opened 2015-09-29
added bridge_ip option

ptinsley:
the default docker subnet was a conflict in our environment. Added a bridge_ip option to make it easy to change that setting.


garethr:
Hi @ptinsley. This looks good, thanks for the addition. Just a couple of points:

* Could you swuash the commits into one commit? It's really just a single logical change.
* Would you be able to add a simple rspec-puppet test? The existing test suite and the details in the CONTRIBUTING doc should help but do feel free to ask questions

ptinsley:
the latest RPMs completely screw up all of this and we've stopped using custom bridge so I'm just going to close this and will revisit later if time allows and the RPM issues get resolved.

PR #355, opened 2015-10-02
Configurable templates for docker::run init.d and systemd.

houtmanj:
Allow the templates for starting an docker container to be replaced.
This would allow us to expand them with actions like loadbalancer updates.


garethr:
@houtmanj thanks for the suggestion, nice idea. Couple of quick fixes hopefully:

* Could you rebase with master so the branch merges cleanly?
* Could you move the parameters to the `docker::run` type? I think they belong there rather than no the main class. This has a few advantages - you don't need the `docker` class (it's not used directly anywhere else in `run` and you can override it on a per container basis rather than just globally.
* It would also be useful to add a quick unit test to demonstrate the functionality, this helps with preventing regressions in the future.

Thanks again. Appreciated. And any questions or help required just let me know.

PR #356, opened 2015-10-02
Ensure that containers are restart after the docker restarts

houtmanj:
If the module decides to reload the docker daemon then all containers are stopped.
This change ensures that all containers are started after the refresh of service.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -26,6 +26,11 @@
 #   (optional) Whether or not to restart the service if the the generated init
 #   script changes.  Default: true
 #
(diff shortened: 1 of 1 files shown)


garethr:
Thanks for submitting. I spotted one issue with the proposed changes unfortunately. The use of `Service['docker']` here couples the `docker::run` type to service, which means you can no-longer use it without also using the module to manage the service. This isn't always required, for instance if you where using it against a remote Swarm or on another platform which ran Docker for you. The module has some prior art, with some in PRs, around disabling the service. It's probably a good idea to piggy-back on that.

PR #357, opened 2015-10-05
New docker repos

rawmannovak:
Changed repo url to use plain HTTP (to enable caching on apt-cacher ng). Also removed requirement for apt-transport-https, since it's not needed anymore (and created dependency loop with puppet 3.7.2)


garethr:
Thanks for submitting, but unfortunately I won't be merging this suggested change. Although HTTPS only adds a little to the signing it does any something. Personally I think it's the right default.

The good news is you can override this default using the `package_source_location` param.

PR #359, opened 2015-10-06
Enabling docker version to be set to present or latest, and option to disable repo management

markasammut:


garethr:
@markasammut the work in master to add support for the new repos changed the code enough here to mean this should should work now in master. I'll close this now but if you still have a problem do open an issue.

PR #360, opened 2015-10-07
Add a new $local_user param to the registry login

kasisnu:
Currently we login to registries as root. This works but breaks the
expectation that anyone in the docker group should be able to use these
registries. This patch exposes a $local_user param which lets us exec as
any user and consumers of this module can specify that in their
manifests.


garethr:
Hi @kasisnu. Thanks for this, much appreciated.

PR #363, opened 2015-10-28
Adding the option to ensure running state of container service

gomex:
That module create a service for each container in "docker::run", but don't ensure that service in running mode. This patch fix it.


garethr:
Hi @gomex. Service has the `running` attribute and the values true and running have the same effect: https://docs.puppetlabs.com/references/latest/type.html#service-attribute-ensure. I'd rather keep the current behaviour as the change would be backwards incompatible. Thanks anyhow.

PR #365, opened 2015-11-03
Use environment variable for registry password.

andygodwin:
I'm using eyaml to encrypt a private docker registry password in a hiera xxx.eyaml file, but if puppet apply is run with --debug, the exec task in docker::registry (auth against ${server}) logs the password in plain text.

This change just puts the password into an environment variable for the exec task, so the logs show '-p ${password}' instead.


garethr:
@andygodwin thanks, thats a nifty pattern. Much appreciated for the PR.

andygodwin:
Thanks @garethr

PR #366, opened 2015-11-04
Fix ensure_packages argument

chrismwendt:
This encloses the argument to `ensure_packages` in an array. Previously, a string was being passed, which results in this error:

```
Error: ensure_packages(): Requires array given (String) at ....../modules/docker/manifests/repos.pp:12
```

Documentation for `ensure_packages`: https://forge.puppetlabs.com/puppetlabs/stdlib#ensure_packages

I'm curious to know whether or not this has been a problem for anyone else too.

### DIFF
--- a/manifests/repos.pp
+++ b/manifests/repos.pp
@@ -9,7 +9,7 @@ class docker::repos {
     'Debian': {
       include apt
       # apt-transport-https is required by the apt to get the sources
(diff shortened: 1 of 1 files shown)


garethr:
@chrismwendt thanks. I think that might have been just broken recently in master. Thanks for catching.

PR #368, opened 2015-11-04
Don't fail the service when the pull doesn't work

jovandeginste:
The image may very well be available locally, and you don't want to
end up with services not starting because of network issues or the like


garethr:
@jovandeginste thanks, nice improvement. Much appreciated.

PR #371, opened 2015-11-09
add tests and docs for storage drivers

robbyt:
related to #339


garethr:
@robbyt thanks, much appreciated.

PR #374, opened 2015-11-12
allow Docker::Image to be refreshed

danielpanteleit:
I introduced another "exec" so that the "$image_install" can be triggered by external resources. The new "exec" checks if the image already exists and triggers the install if necessary. So the image is built if it does not exist or if it is refreshed externally.

With this change I can specify a "Dockerfile" resource that triggers the image on change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 Puppet module for installing, configuring and managing
 [Docker](https://github.com/docker/docker) from the [official repository](http://docs.docker.com/installation/) or alternatively from [EPEL on RedHat](http://docs.docker.io/en/latest/installation/rhel/) based distributions.
- 
+
 [![Puppet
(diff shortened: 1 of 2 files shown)


garethr:
@daniellawrence nice idea. I noticed that this fails the unit tests. If you could update those to describe the behaviour that would be great.

This also sounds like something that would be useful to document in the README. Maybe add a quick example of how you're using it?

And thanks for the suggestion. Much appreciated.

danielpanteleit:
The failing tests were just about the no-op command "true" which seems puppet lint interprets as a quoted boolean value. I changed it to 'echo > /dev/null' and added a paragraph to the README.

danielpanteleit:
@garethr ready to be pulled

Issue #369, opened 2015-11-05
Cyclic dependency caused by apt-transport-https -> Apt::Source <||>


chrismwendt:
When attempting to use the latest version of this Docker module, I encountered a cyclic dependency which is caused by https://github.com/garethr/garethr-docker/blob/master/manifests/repos.pp#L13

```puppet
Package['apt-transport-https'] -> Apt::Source <||>
```

Here is the formatted output of the cycle from Puppet:

```
Anchor[apt::key ----- present] => 
Apt::Key[Add key: ----- from Apt::Source foo] => 
File[foo.list] => 
Exec[apt_update] => 
Class[Apt::Update] => 
Anchor[apt::update] => 
Class[Apt] => 
Package[apt-transport-https] => 
Apt::Source[foo] => 
File[foo.list]
```

Is it absolutely necessary to force all occurrences of `Apt::Source` to depend on `Package['apt-transport-https']`? Perhaps there is a way to specify that it's only required for Docker.

I'm going to throw a bunch of ideas out here in case any of them are viable:

- `if already_has_https { do nothing } else { do this }`
- Pass `https` as a parameter to `Apt::Source` (does `Apt::Source` support this?)
- `include apt-transport-https` instead of introducing the collector syntax `<||>`
- Specify `Apt::Source['docker']` directly instead of using the collector
- Restrict the collector with a search expression inside the collector like `<| title == 'docker' |>`

/cc @kpaulisse

kpaulisse:
I saw this as a chicken-and-egg problem. You need to install `apt-transport-https` in order that `apt-get update` will work with a file that uses https URLs. But you need to have `apt-get update` already run in order to be able to install the package.

I think this is causing a problem for us and possibly not others because we have all package installs dependent on all apt repos, which I think is a reasonable design choice. Specifically I mean that we are already using this:

`Class['apt'] -> Package<||>`

I understand why you are including this code in the docker module, but in my opinion the correct place to fix this is in the `apt` module itself... some sort of "transport packages to install" argument. In the meantime you could consider enclosing the `apt-transport-https` package logic in a conditional statement that can be overridden by a parameter, so that we can use your module even in the presence of the apt to package relationship I noted above.

jkroepke:
```
Class['apt'] -> Package <| title != 'apt-transport-https' |>
```

is a bad workaround..

sigmunau:
This causes puppet to try to install apt-transport-https before installing the apt-source that contains it on our setup.

[...]

PR #382, opened 2015-11-23
install apt-transport-https via required_packages (issue #369)

madddi:
Installing apt-transport-https via required_packages mitigates the cyclic dependency mentioned in issue #369


jkroepke:
required_packages is in apt module deprected and emitts a warning.

manifests/repos.pp
```diff
@@ -7,26 +7,24 @@
...
-        apt::source { 'docker':
-          location          => $docker::package_source_location,
-          release           => $docker::package_release,
-          repos             => $docker::package_repos,
-          key               => $docker::package_key,
-          key_source        => $docker::package_key_source,
-          required_packages => 'debian-keyring debian-archive-keyring',
-          pin               => '10',
-          include_src       => false,
+      if $docker::manage_package {
```
fraenki: +1 Finally this would enable us to manage apt on our own.

garethr:
@madddi thanks for looking at this. I'm having trouble reviewing as it looks like the commit history contains lots of commits from master. If you could rebase against the current master I'll take a closer look and provide feedback. Thanks again.

madddi:
@garethr Sorry, I unfortunately did'nt get around to update this PR. As @jkroepke pointed out, "required_packages" is deprecated in the apt module. Looking at this again I think the best solution would be to provide "manage_repo" and "manage_package" options to disable repository management separately. If you're ok with this I'd like to close this request and open another one with the changes described above.

garethr:
@madddi I think that sounds like a good idea. I'll close now as mentioned. And I'll look out for the PR and try and jump on a review quickly. Thanks

PR #385, opened 2015-11-28
Docker compose

scotty-c:
Adds docker-compose functionality to the module


garethr:
Hi @scotty-c. Thanks for this. I'm working on Compose support in https://github.com/garethr/garethr-docker/pull/385. I think I'm preferring the single binary release for the installation (which seems to be the recommended Docker path for installation).

I'm busy writing tests and will get to documentation later this week, but the current branch seems to work. It does a bit more than your original spike - it only makes compose calls as needed (ie. if the relevant containers aren't present), which helps the type work in an idempotent manner. I'd love any feedback you might have if you get chance to try it out. And thanks again for the original work.

scotty-c:
Good evening Gareth,

No worries I will try it out once you have the realese up.

With the types and providers I wrote for both the networking and compose.
What where the main issues so I know for the future. Just that you had to
re write both instead of working on them as a base.

Kind regards
Scott Coulton

On Tuesday, 26 January 2016, Gareth Rushgrove <notifications@github.com>
wrote:

> Hi @scotty-c <https://github.com/scotty-c>. Thanks for this. I'm working
> on Compose support in #385
> <https://github.com/garethr/garethr-docker/pull/385>. I think I'm
> preferring the single binary release for the installation (which seems to
> be the recommended Docker path for installation).
>
> I'm busy writing tests and will get to documentation later this week, but
> the current branch seems to work. It does a bit more than your original
> spike - it only makes compose calls as needed (ie. if the relevant
> containers aren't present), which helps the type work in an idempotent
> manner. I'd love any feedback you might have if you get chance to try it
> out. And thanks again for the original work.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/garethr/garethr-docker/pull/385#issuecomment-174958549>
> .
>


-- 
Kind regards

Scott Coulton

Sent from Gmail Mobile

garethr:
Hi @scotty-c. Happy to chat through in more detail if we find a time that works for both of us. I've gone over the top on feedback, hopefully that's useful.

Their where a few small things that I'd probably have worked with you (or anyone else on) but worth mentioning just for completeness.

* Missing unit and acceptance tests
* Style (thinks like variable indentation or varying capitalisation in messages). I plan on adding rubocop checks to make this easier to pick up on for contributors
* Opinionated defaults, like the choice of overlay. Better for the end user to be explicit.
* Missing validation in the types
* Missing a few attributes on the network type
* For calling the docker commands you mix called via the `commands` shim and via backticks
* Mixed up usage of properties and params in the types

None of these where blockers in there own right, but add up a little. More important where:

* Although compose is available via pip the main documentation recommends the released packages from GitHub. This has some disadvantages but from the point of view of the modules means we don't have to bring in various python dependencies which felt like a bit of overreach. That would have been a fairly easy replacement.
* The network provider didn't implement `self.instances`. Doing so adds support for using `puppet resource` as an interface to the type, it also avoids the n+1 issue you had by calling `docker network` for every `exists?` call. Adding this changes how the rest of the provider works however.
* The biggest issue with the network interface for me was the use of the `create` param. This is definitely a smell in puppet as you're modelling the desired state, and create is a verb. This design choice I think led to the implementation which was only useful for running once and creating a network - not for running many times in an idempotent manner or to detect drift
* The compose interfaces biggest challenge was the mapping of the location of the params. If you look at my version I simply use the file path as the `namevar`. This removed the need for the two other params at the same time as always providing something that is unique.
* The scale property was also not very explicit. For instance you have to jump between the compose file and the Puppet code to determine which containers you are scaling, and this couples the order in the compose file to the Puppet code in a non-explicit manner.
* The implementation of `exists?` in the compose provider checks for the existence of the compose file, but the intention of the type is to managing the results of compose running against the file (ie. the containers). See the juggling I'm doing with `docker ps` here: https://github.com/garethr/garethr-docker/pull/417/files#diff-bac6fb27ab57e64f8b23753eae68542eR8
* The compose provider also doesn't detect and effect changes, and has an n+1 problem where multiple compose services are managed.

Overall I think the main issues stem from not considering mutation. ie. what happens if you change the code and run it again, and also what happens if the state of the system changes and you run the original code. That's the hard part of developing puppet types and providers I think.

scotty-c:
Good morning Gareth,

Thanks for the feedback.
Just wanted to know to compare your work with mine.
I will have a look at your examples and tidy up one of my modules.

Thanks for taking the time to get back to me. It's greatly appreciated. I
don't often get a chance to work with someone of your level.

Kind regards
Scott

On Wednesday, 27 January 2016, Gareth Rushgrove <notifications@github.com>
wrote:

> Hi @scotty-c <https://github.com/scotty-c>. Happy to chat through in more
> detail if we find a time that works for both of us. I've gone over the top
> on feedback, hopefully that's useful.
>
> Their where a few small things that I'd probably have worked with you (or
> anyone else on) but worth mentioning just for completeness.
>
>    - Missing unit and acceptance tests
>    - Style (thinks like variable indentation or varying capitalisation in
>    messages). I plan on adding rubocop checks to make this easier to pick up
>    on for contributors
>    - Opinionated defaults, like the choice of overlay. Better for the end
>    user to be explicit.
>    - Missing validation in the types
>    - Missing a few attributes on the network type
>    - For calling the docker commands you mix called via the commands shim
>    and via backticks
>    - Mixed up usage of properties and params in the types
>
> None of these where blockers in there own right, but add up a little. More
> important where:
>
>    - Although compose is available via pip the main documentation
>    recommends the released packages from GitHub. This has some disadvantages
>    but from the point of view of the modules means we don't have to bring in
>    various python dependencies which felt like a bit of overreach. That would
>    have been a fairly easy replacement.
>    - The network provider didn't implement self.instances. Doing so adds
>    support for using puppet resource as an interface to the type, it also
>    avoids the n+1 issue you had by calling docker network for every
>    exists? call. Adding this changes how the rest of the provider works
>    however.
>    - The biggest issue with the network interface for me was the use of
>    the create param. This is definitely a smell in puppet as you're
>    modelling the desired state, and create is a verb. This design choice I
>    think led to the implementation which was only useful for running once and
>    creating a network - not for running many times in an idempotent manner or
>    to detect drift
>    - The compose interfaces biggest challenge was the mapping of the
>    location of the params. If you look at my version I simply use the file
>    path as the namevar. This removed the need for the two other params at
>    the same time as always providing something that is unique.
>    - The scale property was also not very explicit. For instance you have
>    to jump between the compose file and the Puppet code to determine which
>    containers you are scaling, and this couples the order in the compose file
>    to the Puppet code in a non-explicit manner.
>    - The implementation of exists? in the compose provider checks for the
>    existence of the compose file, but the intention of the type is to managing
>    the results of compose running against the file (ie. the containers). See
>    the juggling I'm doing with docker ps here:
>    https://github.com/garethr/garethr-docker/pull/417/files#diff-bac6fb27ab57e64f8b23753eae68542eR8
>    - The compose provider also doesn't detect and effect changes, and has
>    an n+1 problem where multiple compose services are managed.
>
> Overall I think the main issues stem from not considering mutation. ie.
> what happens if you change the code and run it again, and also what happens
> if the state of the system changes and you run the original code. That's
> the hard part of developing puppet types and providers I think.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/garethr/garethr-docker/pull/385#issuecomment-175007681>
> .
>


-- 
Kind regards

Scott Coulton

Sent from Gmail Mobile

PR #387, opened 2015-11-30
Fix Docker systemd deprecated cli options

jfarrell:
Fix the systemd daemon flag in debian and rhel templates that became
broken due to the recent Docker upstream commit

https://github.com/docker/docker/commit/7929888214741c4ab194c44e0b14ac08aca06556

### DIFF
--- a/templates/etc/systemd/system/docker.service.d/service-overrides-debian.conf.erb
+++ b/templates/etc/systemd/system/docker.service.d/service-overrides-debian.conf.erb
@@ -2,5 +2,5 @@
 EnvironmentFile=-/etc/default/docker
 EnvironmentFile=-/etc/default/docker-storage
 ExecStart=
-ExecStart=/usr/bin/<%= @docker_command %> -d -H fd:// $OPTIONS \
+ExecStart=/usr/bin/<%= @docker_command %> daemon -H fd:// $OPTIONS \
         $DOCKER_STORAGE_OPTIONS
--- a/templates/etc/systemd/system/docker.service.d/service-overrides-rhel.conf.erb
+++ b/templates/etc/systemd/system/docker.service.d/service-overrides-rhel.conf.erb
@@ -3,7 +3,7 @@ EnvironmentFile=-/etc/sysconfig/docker
 EnvironmentFile=-/etc/sysconfig/docker-storage
 EnvironmentFile=-/etc/sysconfig/docker-network
 ExecStart=
-ExecStart=/usr/bin/<%= @docker_command %> -d -H fd:// $OPTIONS \
+ExecStart=/usr/bin/<%= @docker_command %> daemon -H fd:// $OPTIONS \
       $DOCKER_STORAGE_OPTIONS \
       $DOCKER_NETWORK_OPTIONS \
       $BLOCK_REGISTRY \


garethr:
Mmm, do you know for what versions of docker this is invalid? Wondering when the daemon option was added with regards backwards compatability for the module, and whether this needs to be clever.

jfarrell:
- 1.9 removed, `docker daemon` only supported
- 1.8 deprecated notice added, occurred during the cli rewrite [1] supports both `docker -d` and `docker daemon`

[1]: https://github.com/docker/docker/commit/96ce3a194aab2807fdd638825b9ea7cb9ba55c36

garethr:
Mmm, so `docker daemon` appears to have only been added in 1.8 (http://www.projectatomic.io/blog/2015/09/docker-daemon-command/). Lots of users of the module are not using recent versions so I'm hesitant to merge this as is, as it prevents folks upgrading.

jfarrell:
easy enough to add a check in the template if the version is < 1.8 use `-d` flag, update for this coming shortly

garethr:
So, it looks like this wasn't removed in 1.9, https://twitter.com/calavera/status/672457562051117056, but is still deprecated.

jfarrell:
hmm, was seeing an issue with systemd and the -d option after the 1.9.1 update, looking at the code though the -d option is still available with the deprecated notice. Also digging into the patch for this not sure that there is an easy route as I initially thought, cant use $version as it can be undefined and could use a custom fact to check the version, but would have to parse `docker -v` as the custom package name param prevents using the default package name with something like

```
case Facter.value('osfamily')
    when "RedHat"
      version = Facter::Util::Resolution.exec('/bin/rpm -q --queryformat "%{VERSION}-%{RELEASE}" docker-engine')
    when "Debian"
      version = Facter::Util::Resolution.exec('/usr/bin/dpkg-query -W -f=\'${Version}\' docker-engine')
    else
      nil
    end
```

thoughts on which avenue you would like to go with this?

gdhbashton:
Just a thought. One of the new features in 1.10 is `docker update`. This command did not exist in 1.8 or 1.9, and hence `docker update --help` will have a returncode of zero on 1.10, and one on 1.8 and 1.9:

```
[root@dockerhost ~]# docker version
Client:
 Version:      1.8.3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   f4bf5c7
 Built:        Mon Oct 12 06:06:01 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      1.8.3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   f4bf5c7
 Built:        Mon Oct 12 06:06:01 UTC 2015
 OS/Arch:      linux/amd64
[root@dockerhost ~]# docker update --help
docker: 'update' is not a docker command.
See 'docker --help'.
[root@dockerhost ~]# echo $?
1
```

```
[root@dockerhost ~]# docker version
Client:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 18:34:50 2016
 OS/Arch:      linux/amd64
Cannot connect to the Docker daemon. Is the docker daemon running on this host?
[root@dockerhost ~]# docker update --help

Usage:	docker update [OPTIONS] CONTAINER [CONTAINER...]

Update resources of one or more containers

  --blkio-weight          Block IO (relative weight), between 10 and 1000
  --cpu-shares            CPU shares (relative weight)
  --cpu-period            Limit CPU CFS (Completely Fair Scheduler) period
  --cpu-quota             Limit CPU CFS (Completely Fair Scheduler) quota
  --cpuset-cpus           CPUs in which to allow execution (0-3, 0,1)
  --cpuset-mems           MEMs in which to allow execution (0-3, 0,1)
  --help                  Print usage
  --kernel-memory         Kernel memory limit
  -m, --memory            Memory limit
  --memory-reservation    Memory soft limit
  --memory-swap           Swap limit equal to memory plus swap: '-1' to enable unlimited swap
[root@dockerhost ~]# echo $?
0
```

nickchappell:
This PR's fix worked for me on EL7 and Docker 1.10.0.

garethr:
Closing in favour of #428, which makes this configurable at the top level. It might be worth doing the route @gdhbashton suggested on top of that at a later date.

PR #388, opened 2015-12-02
DRY-ing out docker::service

danzilio:


danzilio:
I feel like a lot of this should be parameterized with most of this logic being moved into `docker::params`

garethr:
@danzilio thanks, and agreed. This looks good. When I have a moment I'll run the acceptance tests and merge. 

Regarding moving things into params, agreed. This is mainly a result of many hands and light work. If you fancy running at that please do.

flyinprogrammer:
THANK YOU FOR THIS WORK!!! you are both amazing!!!

danzilio:
@flyinprogrammer you're very welcome and thank you!

PR #393, opened 2015-12-14
Extend service dependencies

jovandeginste:
Cleanup dependencies:

* add extra parameter for non-docker-dependencies
* add 'after' parallel to 'depends' (includes and obsoletes #317)
  * after only implies order
  * depends implies two-way dependency
* calculate arrays beforehand and clean the code


garethr:
@jovandeginste thanks for this, nice work. Thanks in particular for adding the tests and updating the documentation - much appreciated.

PR #395, opened 2015-12-17
Add new parameters for docker daemon

rchicoli:
Implemented some options for docker daemon 

* bind the docker daemon to multiple ipaddress (changed string to array)
* add ip_forward, iptables, ip_masq, fixed_cidr and bridge option

```hiera
docker::tcp_bind:
  - 'tcp://127.0.0.1:2375'
  - "tcp://%{::ipaddress}:2375"
docker::ip_forward: false
docker::iptables: false
docker::ip_masq: false
docker::fixed_cidr: '10.10.1.0/24'
docker::bridge: 'br0'
docker::default_gateway: '10.10.0.1'
```

OPTIONS=" -H tcp://127.0.0.1:2375 -H tcp://10.10.0.3:2375 --ip-masq=false --iptables=false --ip-forward=false --fixed-cidr 10.10.1.0/24 --default-gateway 10.10.0.1 --bridge br0"


garethr:
Thanks @rchicoli, this looks great, one quick suggestion and one request.

Rather than break backwards compatibility with the change from a string to an array, it's probably worth calling `any2array` on the passed value. You can see an example of this in use here:

https://github.com/garethr/garethr-docker/blob/543b66fd58676164f5a6f7aab99508b440777346/manifests/run.pp#L128

This maintains the ability to pass a string, but allows for passing an array, without having to change the template (which always gets an array).

Also, would it be possible to add some quick unit tests to verify the changes? You should find some other examples but if you have any questions I'm happy to help. The tests really help with ensuring future changes don't break things.

Thanks again.

rchicoli:
It sounds good.
I've already implemented any2array function, but now I am looking up how to get beaker-rspec running on my computer.

At the moment I am facing the following problems when I try to run ```BEAKER_setfile=spec/acceptance/nodesets/debian-81-x64.yml rake beaker```

```ruby
/usr/local/rvm/gems/ruby-2.2.3@3823/gems/beaker-2.32.1/lib/beaker/hypervisor/vagrant.rb:196:in
 `block (2 levels) in vagrant_cmd': Failed to exec 'vagrant destroy --force'. 
Error was /usr/share/gems/gems/bundler-1.7.8/lib/bundler/spec_set.rb:92:in `block in materialize':
 Could not find rake-10.5.0 in any of the sources (Bundler::GemNotFound) (RuntimeError)
```

So I might push the changes changes tomorrow, after solving this little problem

garethr:
@rchicoli this looks like you be not have the bundle correctly setup or up-to-date. Here's a few things to try:

```
BEAKER_setfile=spec/acceptance/nodesets/debian-81-x64.yml bundle exec rake beaker
```

Here you're explicitly running rake with bundler. This avoids issues with rake being on the path and not picking up the relevant bundle.

If that doesn't work then first try:

```
bundle install
bundle update
```

That should install the relevant gems, or update them if they are out of date and that's causing the issue. Let me know if any of that fixes the issues.

rchicoli:
Thanks @garethr, it is working now, but for some reasons I had to upgrade vagrant to 1.8.1 and use ruby 2.1.6.
I am running the tests now. When this is finished I'll push the changes.

rchicoli:
Although it was cool playing around with beaker for a full test environment, I've decided check only the parameters passed to the docker daemon.

So the problem I was having was when I run docker_full_spec, it reads the first lines and enables ip_forward and set default docker iptables

```ruby
19       let(:pp) {"
20         class { 'docker': }
21       "}
```

Then add some context to the same describe section like

```ruby
context 'disable docker\'s addition of iptables rules' do
  before(:all) do
    @pp =<<-EOS
      class { 'docker':
        iptables => false,
      }
    EOS
    apply_manifest(@pp, :catch_failures => true)
    # A sleep to give docker time to execute properly
    sleep 4
  end 

  it 'should run idempotently' do
    apply_manifest(@pp, :catch_changes => true) unless fact('selinux') == 'true'
  end 

  it 'should not enable iptables' do
    shell('iptables -L DOCKER') do |r| 
      expect(r.stdout).to match(/0/)
    end
  end 
end
```

But once ip_forward and iptables have been enabled,  it won't delete iptables DOCKER chain by itself after the daemon reloads with the new options.
I don't know, if you have better idea how to test it, feel free to point it out.

The result should not contain "Chain DOCKER"

```ruby
ubuntu-1404-x64 14:33:07$ iptables -L DOCKER
  Chain DOCKER (1 references)
  target     prot opt source               destination
```

Anyways I think that the spec checks should be enough. What do you think?

garethr:
@rchicoli yeah, I've solved that in other cases with matrix builds and different entry points, but for this test case and this feature I'm happy with your suggestion of the spec tests. Thanks

rchicoli:
hey @garethr, good to know. I will keep that in mind.
Anyways I'm glad to contribute. The code has been rebased and I hope it's not a big deal the commit messages, because I didn't feel like rewriting the history today. Tell me if it is fine. Thanks and keep up the good work.

rchicoli:
Btw I think the notify attribute on service.pp file should be changed to ```notify => $_manage_service,``` instead of ```Service['docker']```

https://travis-ci.org/garethr/garethr-docker/jobs/107876029
https://github.com/rchicoli/puppet-docker/blob/systemd-options/manifests/service.pp#L183

```puppet
if $_service_config {
  file { $_service_config:
    ensure  => present,
    force   => true,
    content => template($service_config_template),
    notify  => Service['docker'],
  }
}
```
don't you think so?
this is not related to this PR, so should I create a new one or do you create the hotfix by youself?

garethr:
Thanks, merged. And the issue with `Service` is also fixed in master. Thanks!

PR #396, opened 2015-12-18
add service_manage parameter

rchicoli:
### service_manage

Specify whether the service should be managed. Valid values are 'true', 'false'. Defaults to 'true'.

```yaml
docker::service_manage: false
```
Since docker doesn't provide a reload option. I would use this option for production environments. Otherwise all docker containers running on the host will be shutdown, when the docker daemon gets reloaded.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -98,6 +98,11 @@
 #   Whether you want to docker daemon to start up at boot
 #   Defaults to true
 #
+# [*service_manage*]
+#   Specify whether the service should be managed.
+#   Valid values are 'true', 'false'.
+#   Defaults to 'true'.
+#
 # [*root_dir*]
 #   Custom root directory for containers
 #   Defaults to undefined
@@ -253,6 +258,7 @@ class docker(
   $package_key_source                = $docker::params::package_key_source,
   $service_state                     = $docker::params::service_state,
   $service_enable                    = $docker::params::service_enable,
+  $service_manage                    = $docker::params::service_manage,
   $root_dir                          = $docker::params::root_dir,
   $tmp_dir                           = $docker::params::tmp_dir,
   $manage_kernel                     = $docker::params::manage_kernel,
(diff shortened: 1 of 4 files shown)


garethr:
@rchicoli thanks for this, nice idea. A prior PR just tidied up the service manifest which is good as it should avoid some of the duplication you had to deal this, but the downside is this PR needs a quick rebase against master to merge cleanly.

It would also be useful to add a quick unit test to demonstrate that, when passing `false` to the new param that the service isn't present in the catalogue.

Any questions, or help with the above, do let me know.

rchicoli:
thanks for the advice. I've run a quick rebase against the master and also add a unit test for service_manage param.
Please let me know, If I forgot something else.

rchicoli:
I am not quite sure why the checks are failing. It should have nothing to do with my changes.

```ruby
should contain File[/etc/systemd/system/docker-sample.service] with content =~ /Restart=on-failure/ (FAILED - 1)

Failure/Error: it { should contain_file(initscript).with_content(/Restart=on-failure/) }
NoMethodError:
undefined method `resource' for nil:NilClass
```
Could you please have a look?
Thanks

garethr:
@rchicoli thanks, I came across this elsewhere this morning. Tracking it down now. It may be related to the Puppet 3.8.5 release yesterday, or something else upstream.

garethr:
@rchicoli that was fun. I tracked this down to a fight between rspec-puppet 2.3.0 and the new Puppet 3.8.5 release. For the moment I'm pinning to 2.2.0. Full details with entertaining test case at https://github.com/rodjek/rspec-puppet/issues/353.

If you rebase on master this test failure should disappear.

rchicoli:
So far it looks great. Good job!

rchicoli:
there is another weird problem

```ruby
Bundler::GemspecError: Could not read gem at /home/travis/.rvm/gems/ruby-2.1.0/cache/json_pure-1.8.3.gem. It may be corrupted.
An error occurred while installing json_pure (1.8.3), and Bundler cannot
continue.
Make sure that `gem install json_pure -v '1.8.3'` succeeds before bundling.
```

could you please check it out?
Thanks

garethr:
@rchicoli it was a transient issue with Travis. Rerunning that job a second time passed. Thanks

garethr:
@rchicoli Thanks! I've merged this in with one change, I've swapped the name to `manage_service` to match the other `manage_*` params. 

There are a few inconsistencies in naming to be sure. I'm on a feature splurge at the moment, but planning a bit of a refactor splurge after that with a focus on consistency, deprecations and tidyup after so many small changes.

PR #397, opened 2015-12-22
Do less if upstream repo is disabled

dvorak:
If the upstream repo management is disabled, then we really shouldn't be
doing anything to manage upstream repos.  Moved apt deps and includes
into the conditional to prevent issues with how the local install wants
to manage repos.

### DIFF
--- a/manifests/repos.pp
+++ b/manifests/repos.pp
@@ -7,21 +7,16 @@ class docker::repos {
 
   case $::osfamily {
     'Debian': {
-      include apt
-      if $::operatingsystem == 'Debian' and $::lsbdistcodename == 'wheezy' {
-        include apt::backports
-      }
-      if ($docker::docker_cs) {
-        $location = $docker::package_cs_source_location
-        $key_source = $docker::package_cs_key_source
-        $package_key = $docker::package_cs_key
-      } else {
-        $location = $docker::package_source_location
-        $key_source = $docker::package_key_source
-        $package_key = $docker::package_key
-      }
-      Exec['apt_update'] -> Package[$docker::prerequired_packages]
       if ($docker::use_upstream_package_source) {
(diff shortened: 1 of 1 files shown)


manifests/repos.pp
```diff
@@ -7,15 +7,16 @@
...
     'Debian': {
-      include apt
-      # apt-transport-https is required by the apt to get the sources
-      ensure_packages(['apt-transport-https'])
-      Package['apt-transport-https'] -> Apt::Source <||>
-      if $::operatingsystem == 'Debian' and $::lsbdistcodename == 'wheezy' {
-        include apt::backports
-      }
-      Exec['apt_update'] -> Package[$docker::prerequired_packages]
       if ($docker::use_upstream_package_source) {
```
fraenki: Wouldn't it be a better approach to check for `if $docker::manage_package` instead? At least this is what I've done locally and what can be found in many other puppet modules as well.

fraenki:
:+1: because I want to get rid of my local modifications

dvorak:
Personally I think including apt and managing all of the things it needs it gross in general, but I figured just moving them inside the conditional was less of an opinionated change.

garethr:
@dvorak @fraenki I'd be happy for this to jump inside a `manage_package` clause I think.

fraenki:
:+1:

madddi:
:+1:

dvorak:
Updated to move inside of existing manage_package block.

dvorak:
This fixes #503 and will pass CI again when #548 is resolved.

dvorak:
I rebased this and it's passing CI again.  I believe the previous concerns have been addressed.  @garethr is there anything I need to do with this?

garethr:
@dvorak thanks for tidying that up, appreciated.

PR #398, opened 2015-12-23
Bug/systemd reload order

wscheele:
we ran into a very annoying issue where docker wouldn't start because of the way the systemd service definition was managed.
initially the service was started without '/etc/systemd/system/docker.service.d/service-overrides.conf'. so would start with a default loopback dm setup in /var/lib/docker/devicemapper.
we had configured a non loopback dm storage, which caused the docker daemon to error on the systemd-reload and restart after the service-overrides was written. lvm setup got messed up, and system state fairly unusable.

this pr addresses the issue we faced by seperating the systemd-reload logic for initial service definition from the reload logic used by docker::run, and introducing stricter dependencies between the puppet resources. we were no longer able to reproduce the issue with these changes.

also note that the issue was only reproducible with a manifest that combined the docker service and some container startups with docker::run. with a simple "include 'docker'", the docker config was being setup correctly.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -58,12 +58,14 @@ class docker::params {
           $package_release = "ubuntu-${::lsbdistcodename}"
           if (versioncmp($::operatingsystemrelease, '15.04') >= 0) {
             include docker::systemd_reload
+            include docker::systemd_reload_before_service
           }
         }
         default: {
           $package_release = "debian-${::lsbdistcodename}"
           if (versioncmp($::operatingsystemmajrelease, '8') >= 0) {
             include docker::systemd_reload
+            include docker::systemd_reload_before_service
           }
         }
       }
@@ -128,6 +130,7 @@ class docker::params {
           $docker_group = 'dockerroot'
         }
         include docker::systemd_reload
+        include docker::systemd_reload_before_service
       }
 
       # repo_opt to specify install_options for docker package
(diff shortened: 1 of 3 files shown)


garethr:
Thanks for the detailed report. This might be the same issue reported in https://github.com/garethr/garethr-docker/issues/407.

I noticed that the unit tests are currently failing on this PR? Due to Puppet lint style checks it looks like: https://travis-ci.org/garethr/garethr-docker/jobs/98958052. Any help required or questions let me know.

It might be worth squashing the commits as well? The diff is quite small but spans 19 commits.

If you could fix up the tests I'll get this merged and released, thanks again for reporting and fixing the issue.

wscheele:
#407 does read as if it's the same issue.
i merged master back ontop, and attempted to rebase, but history on this branch is a bit odd, i got a bit lost in the merge conflicts. will await next checks to see if there's still lint issues (i thought i had resolved these already). can you live with the large commit span?
I'll see if I can easily retest this somehow as the service.pp file has changed rather drastically on master.

garethr:
@wscheele thanks. This looks good. A few of the tests fail around dependency cycles. If you have time to fix up that would be great, otherwise I can take a closer look.

jfarrell:
great work on this @wscheele, I stumbled onto the same issue using centos 7.2 and can confirm your patch does address the issue. I tracked down why the tests where throwing a dependency cycle and fixed it, combined patch set in pr #424

PR #402, opened 2016-01-07
Remove container with its volumes

jfroche:
We should get rid of the volumes associated with the container we want
to delete (not using the `docker volume` api yet).


garethr:
@jfroche thanks, good suggestion. Appreciated.

PR #403, opened 2016-01-07
Enable ensure => absent parameter

jfroche:
to:
 - remove container (with its volumes)
 - disable the service
 - remove the init script


garethr:
@jfroche thanks, much appreciated. An oft-requested feature. Thanks in particular for the acceptance tests work, it makes changes like this much easier to verify and merge. Cheers again.

PR #410, opened 2016-01-19
ensure idempotence

binford2k:
This ensure idempotence when multiple `docker::run` resources are declared


garethr:
@binford2k ooh, thanks. Appreciated.

PR #411, opened 2016-01-20
Add before_start parameter for docker::run

dvorak:
This adds a complement to before_stop.  Any command passed in as the
before_start parameter will be run before beginning any start related
cleanup or commands.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -83,6 +83,7 @@ define docker::run(
   $socket_connect = [],
   $hostentries = [],
   $restart = undef,
+  $before_start = false,
   $before_stop = false,
 ) {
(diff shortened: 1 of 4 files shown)


garethr:
Thanks, good idea. Could you add a quick unit test or two. The tests really help with not breaking functionality in the future. You can likely follow the same pattern as the `before_stop` tests at https://github.com/garethr/garethr-docker/blob/master/spec/defines/run_spec.rb#L328-L336.

Any questions or help required just let me know. And thanks again.

dvorak:
Sure, no problem.

garethr:
@dvorak thanks. Appreciated.

PR #412, opened 2016-01-21
Define service_hasstatus service_hasrestart parameters

jfroche:
These two parameters weren't set in different OS family/system. systemd enabled OS have restart and status capability.


garethr:
@jfroche thanks! Appreciated. Especially for adding the quick test. Really helps with all the changes from different folks.

PR #413, opened 2016-01-21
Update run manifest to use service_prefix and docker_command params throughout

jfarrell:
docker_command and service_prefix are missed in a couple locations, this updates the run manifest to use them through out


garethr:
@jfarrell ah, much appreciated. Thanks for catching those.

PR #414, opened 2016-01-22
Change update_docker_image from File to ensure_resource

jfarrell:
Changes how update_docker_image.sh is created and adds an ensure so it can
properly be removed if ensure => absent is set. Also adds test case to cover the
missed resource in the coverage report.

Introduced in pr #237

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -35,6 +35,17 @@ define docker::image(
   validate_re($image, '^[\S]*$')
   validate_bool($force)
 
+  # Wrapper used to ensure images are up to date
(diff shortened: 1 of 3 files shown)


jfarrell:
single travis failed run due to bundler issue not related to this patch

```
Bundler::GemspecError: Could not read gem at /home/travis/.rvm/gems/ruby-2.1.0/cache/puppet-4.0.0.gem. It may be corrupted.
An error occurred while installing puppet (4.0.0), and Bundler cannot continue.
```

garethr:
@jfarrell thanks, good idea. Appreciated.

PR #418, opened 2016-01-26
Amazon linux is still based on centos 6.x, yet the version is like 

b2jrock:
'2015.xx'. It still requires detach in init as far as I can tell. This seems to have regressed somewhere

I didn't see a good place to insert a test since it'd only be able to run in AWS, but would happy to take a pointer and run with it.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -147,7 +147,7 @@ class docker::params {
       $package_release = undef
       $service_name = $service_name_default
       $docker_command = $docker_command_default
-      if versioncmp($::operatingsystemrelease, '7.0') < 0 {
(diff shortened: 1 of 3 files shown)


garethr:
@b2jrock thanks for the fix. The best place to test this would be in the unit tests I think. I'm happy to merge without, but the test would help it avoid being broken again. It's hard to test Amazon Linux in the current acceptance tests unfortunately.

b2jrock:
I added Amazon to the run_spec. It seems slightly heavy weight, but Amazon linux is different enough that having a full set of tests at this level may be the only way to ensure that new changes don't create regressions.

garethr:
@b2jrock thanks, appreciated.

PR #419, opened 2016-01-26
Add MIN_DATA_SIZE to docker-storage-setup template

jfarrell:
Add a storage template configuration option MIN_DATA_SIZE to allows users to specify minimum size of data pool volumes. This is needed to prevent small pools from hanging on the first image pull.


garethr:
Thanks @jfarrell. Appreciated.

PR #421, opened 2016-01-28
Add dm_option for deferred device deletion

jfarrell:
Adds the missing option for deferred device deletion to storage configs

  --storage-opt dm.use_deferred_deletion=true

Introduced in Docker pr 16381


garethr:
@jfarrell thanks!

PR #423, opened 2016-01-30
Chained ordering leaves out Docker::Registry

### LINKED ISSUE #408: Chained ordering leaves out Docker::Registry
We started using the new class docker::registry.  However, there's nothing imposing that the login exec is called before calling docker::image and the docker pull exec.

If done out of order, the image pulls from a secure registry may fail.

jfarrell:
Updates ordering to ensure Registries are created before Image or
Run resources could potentially try and use them.

Closes #408


garethr:
Thanks @jfarrell, appreciated.

PR #424, opened 2016-01-30
prevent docker service startup with wrong configuration

### LINKED ISSUE #407: Missing dependency on service override file and systemd reload
I am running Docker 1.8.3 on CentOS 7.

I am specifying multiple parameters for starting the Docker daemon such as DNS and storage options. In the new version of the Puppet module there is a bug causing these options to not load correctly.

The problem is in Docker::Service where the service-overrides files are templated. These files notify `docker-systemd-reload`, but there is no dependency to insure that the reload occurs before the docker service is started. To be honest, there is not even a dependency insuring that the file is present before the service is started.

In my deployment this caused the service to be started using the default installation service file with none of my custom options. The file was templated and/or loaded after the service was started. This means a restart of the service would include my custom options, but unfortunately the specific options I was using (dm.thinpooldev) caused the restart to fail.

Because Docker::Service is using the same `docker-systemd-reload` exec that Docker::Run uses for individual containers, I cannot see a way to resolve this issue without changes to the module. I cannot impose a dependency in my code that uses `docker-systemd-reload` as it will affect all containers in addition to the service. So something like `Exec['docker-systemd-reload'] -> Service['docker']` will create a dependency cycle.

The only immediate solution that came to mind is to have the service notify a differently named but otherwise identical execution resource (`service-systemd-reload`?) so that an appropriate dependency can be introduced for the service only.

While I am on CentOS, from looking at the code I believe the problem exists on Debian and ArchLinux as well.

jfarrell:
Intitial patch submitted by @wscheele in #398

This patch extends the work initially done by rebasing against current master
and fixing the broken test case dependency cycles.

Closes #398
Closes #407

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -67,6 +67,7 @@ class docker::params {
             $service_hasstatus       = true
             $service_hasrestart      = true
             include docker::systemd_reload
+            include docker::systemd_reload_before_service
           } else {
             $service_config_template = 'docker/etc/default/docker.erb'
             $service_provider        = 'upstart'
@@ -84,6 +85,7 @@ class docker::params {
             $service_hasstatus       = true
             $service_hasrestart      = true
             include docker::systemd_reload
+            include docker::systemd_reload_before_service
           } else {
             $service_config_template = 'docker/etc/default/docker.erb'
           }
@@ -163,6 +165,7 @@ class docker::params {
           $docker_group = 'dockerroot'
         }
         include docker::systemd_reload
+        include docker::systemd_reload_before_service
       }
 
       # repo_opt to specify install_options for docker package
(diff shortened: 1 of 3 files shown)


jfarrell:
1/3 of the travis runs failed with the dependency cycle issue, looking into what the difference is

garethr:
@jfarrell looks like an issue only on the old parser, and then only on Arch. If you get stuck shout and I'll jump in and have a look too.

garethr:
@jfarrell thanks for the work on this, and @wscheele too. I made a couple of fixes and have this in master now. The failing tests actually showed up a confusion about the service_name param, using it direct from params prevents it being overridden. Thanks all!

PR #425, opened 2016-02-02
Add description to yum repo

jfarrell:
Adds a description to yumrepo to remove warnings of "Repository 'docker' is missing name in configuration, using id".


garethr:
Thanks @jfarrell

PR #428, opened 2016-02-04
Use current docker daemon command

samuelson:
`docker -d` has been deprecated
the command is now `docker daemon`
https://github.com/docker/docker/blob/master/CHANGELOG.md#180-2015-08-11

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -204,6 +204,10 @@
 #   Specify a custom docker command name
 #   Default is set on a per system basis in docker::params
 #
+# [*daemon_flag*]
+#  Specify a flag for running docker as daemon
+#  Default is set on a per system basis in docker::params
+#
 # [*docker_users*]
 #   Specify an array of users to add to the docker group
 #   Default is empty
@@ -291,6 +295,7 @@ class docker(
   $package_name                      = $docker::params::package_name,
   $service_name                      = $docker::params::service_name,
   $docker_command                    = $docker::params::docker_command,
+  $daemon_flag                       = $docker::params::daemon_flag,
   $docker_users                      = [],
   $repo_opt                          = $docker::params::repo_opt,
   $nowarn_kernel                     = $docker::params::nowarn_kernel,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -43,6 +43,7 @@ class docker::params {
   $service_name_default              = 'docker'
   $docker_command_default            = 'docker'
   $docker_group_default              = 'docker'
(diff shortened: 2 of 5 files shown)


garethr:
There is some discussion of this in #395. The issue is that this was added in 1.8, so won't work for anyone using docker versions prior to that. I have anecdotal evidence that's a non-zero group of folks. I'm happy running deprecated options, I've not verified whether this is removed in 1.10 yet however.

gdhbashton:
New CentOS 7 VM, freshly provisioned today so it's grabbed the latest version of Docker (currently `1.10.0` from https://yum.dockerproject.org/repo/main/centos/7/Packages/ )

```Feb 05 10:23:01 dockerhost systemd[1]: Starting Docker Application Container Engine...
Feb 05 10:23:01 dockerhost docker[12855]: flag provided but not defined: -d
Feb 05 10:23:01 dockerhost docker[12855]: See '/usr/bin/docker --help'.
Feb 05 10:23:01 dockerhost systemd[1]: docker.service: main process exited, code=exited, status=125/n/a
Feb 05 10:23:01 dockerhost systemd[1]: Failed to start Docker Application Container Engine.
Feb 05 10:23:01 dockerhost systemd[1]: Unit docker.service entered failed state.
Feb 05 10:23:01 dockerhost systemd[1]: docker.service failed.
```

jfarrell:
@gdhbashton I hit the same issue you saw with centos and had opened #387, no ideal approach to support both `-d` and `daemon` options

AlexRRR:
>  I've not verified whether this is removed in 1.10 yet however.

I can confirm it is removed now.

samuelson:
Yeah, that's why I filed the issue/PR it broke on my Centos7 VM, the -d option is removed in 1.10.  It looks like they only left both options in for 1.9.

Since the systemd file is a template already, it shouldn't be too tough to use different versions, at very least to add a parameter.

samuelson:
Ok, that ^^ fixes merge conflict and adds a parameter for setting the flag.

samuelson:
@garethr let me know if you'd like me to fix that in a different way.

garethr:
@samuelson thanks for this, much appreciated. I've made a minor change, changing the name from `_flag` to `_subcommand`. Going from `-d` (a flag) to `daemon` (subcommand) makes this confusing either way, but I reason better fall forward than fall backwards. Thanks again.

jfarrell:
#355 might be the final part to this to allow for complete flexibility by overriding the full template if someone needed without having to introduce a ton of new feature flags

PR #429, opened 2016-02-05
fix docker network type and readme

hdeadman:
The readme has the wrong attribute name iprange should be ip_range and I got this error when running on centos7:

Execution of '/usr/bin/docker network create --driver overlay --subnet
192.168.1.0/24 --gateway 192.168.1.1 --ip-range 192.168.1.4/24 adnets'
returned 1: flag provided but not defined: --driver overlay
See '/usr/bin/docker network create --help'.
flag provided but not defined: --driver overlay

So I put = between the args and the value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,11 +316,11 @@ like:
 
 ```puppet
 docker_network { 'my-net':
-  ensure  => present,
-  driver  => 'overlay',
-  subnet  => '192.168.1.0/24',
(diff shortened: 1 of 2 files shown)


garethr:
@hdeadman thanks for diving in and checking out the new network feature, and thanks for the fixes. Much appreciated.

PR #431, opened 2016-02-09
Fix services overrides template daemon command

jfarrell:
params daemon_command_default is never set to the global daemon_command which
causes the service overrides templates to have a blank value in the erb
ExecStart line.


garethr:
Thanks, I'd spotted that this morning as well while doing some more testing.

PR #433, opened 2016-02-09
Notify service only if the _manage_service is not undef.

mterzo:


jfarrell:
+1, nice catch @mterzo

PR #436, opened 2016-02-11
Allow to turn of managing `apt` class.

aequitas:


garethr:
The module already has `manage_package` which accomplishes the same thing for the RHEL line, I think rather than add a new param this would be better reusing that one, but expanding it's scope to not just deal with the relationship but the whole repo.

aequitas:
thx for the quick reply, fixed.

PR #437, opened 2016-02-12
Allow overwriting repo_opt with ''

felixb:
Setting it to `undef` makes puppet use the default from `::docker::params`.
This change allows us to set it to `''` instread omit any additional `install_options`.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -69,10 +69,10 @@ class docker::install {
 
   if $docker::manage_package {
 
-    if $docker::repo_opt {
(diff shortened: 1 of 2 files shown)


jfarrell:
+1, looks good

PierreR:
+1 
Actually the default `repo_opt' params won't work in rhel7 and this patch is the only way out of it (at least from my personal experience).

garethr:
@felixb thanks for the fix. I'll have a think about a nicer interface for this, but for now this solves the problem. Much appreciated.

PR #441, opened 2016-02-12
Add manager flag for docker-storage-setup

jfarrell:
Fixes a race condition with docker-storage-setup sysconfig template, the
package and its service. Docker-storage-setup depends on sysconfig/docker-storage
and sysconfig/docker-storage-setup files to be in place before it runs. By default
it runs as a service before the docker service starts. Currently there is an
ordering issue because the docker-storage-setup service is not being managed by
the docker::service class but the sysconfig file is, this causes the
docker-storage-service to not startup.


garethr:
Thanks. Travis hit issues for this, although again only under some runtimes.

```
The parameter 'storage_config' is declared more than once in the parameter list at /home/travis/build/garethr/garethr-docker/manifests/service.pp:97:3
The parameter 'storage_config_template' is declared more than once in the parameter list at /home/travis/build/garethr/garethr-docker/manifests/service.pp:98:3
```

jfarrell:
sorry @garethr, missed that in the rebase. fixed now

jfarrell:
@garethr, patch updated against master and all tests passing

eperdeme:
How come this did not get merged in? It seems required when I look at it as unless I reboot after the puppet run docker-storage-setup never starts

jfarrell:
we where using this patch in production for months and can confirm that it does take care of the issue you are talking about. We have now switched over to using overlay and since this had not been merged and fell through the cracks I closed it due to no longer needing it and would not be rebasing it to keep current. Please feel free to take it over, rebase and resubmit the patch

jfarrell:
docker-storage-setup should also not be required in install.pp as a default for rhel systems if this flag is not in use, but thats a minor detail that had been missed in my initial pull request

Issue #401, opened 2016-01-05
resource cycles


bulemoon:
When I followed the step on https://github.com/garethr/puppet-docker-swarm-example. It had ''Resource Cycles" issue as following.
digraph Resource_Cycles {
  label = "Resource Cycles"
"Exec[docker-systemd-reload]" -> "Service[docker-swarm]" -> "Docker::Run[swarm]" -> "Docker::Run[swarm-manager]" -> "File[/etc/systemd/system/docker-swarm-manager.service]" -> "Exec[docker-systemd-reload]"
}

I just run swarm-manager and swarm and don't use docker class.

PR #443, opened 2016-02-16
Fix dependency cycle with docker::run

ayohrling:
The docker::run command when using systemd notifies the exec for
reload, but then the service requires this exec. If you have multiple
runs, this creates a dependency cycle. I added a new exec based on the
sanitised title of the run so sure it wastes cycles reloading multiple
times, but it doesn't fail.

This addresses #401

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -326,8 +326,14 @@ define docker::run(
         }
       }
       if $uses_systemd {
-        File[$initscript] ~> Exec['docker-systemd-reload']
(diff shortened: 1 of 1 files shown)


garethr:
@ayohrling thanks, good sleuthing.

zhifengMaBeijing:
Hi garethr, 

which version will deliver this fix?

regards,
Zhifeng

garethr:
@zhifengMaBeijing this has been released in 5.2.0

PR #445, opened 2016-02-21
allow to define default extra_systemd_parameters

madddi:
Allows to define systemd unit parameters for all docker::run defines.


garethr:
@madddi mmm, I'm not sure about this. It couples the `run` resource to the `docker` class. At the moment you can use `run` independently. I wonder if a better approach is resource defaults? Worth having a think about.

madddi:
@garethr Yes, until recently I wasn't aware of resource defaults.

Let's keep `docker::run`independent.

PR #446, opened 2016-02-23
add docker stop with stop timeout

lkosz:
I've made two changes in module:
- add support of timeout in docker stop: docker stop -t <timeout> (optional)
https://docs.docker.com/engine/reference/commandline/stop/

- one correction of typo in CONTRIBUTING.md

I added one test which checks if docker stop -t <timeout> is added. More detailed tests could check validation of parameter and it initial state. If you want them, let me know :)

~~I found one problem - there is difference between init script and systemd service in stop methods: systemd makes docker kill, init script makes docker stop. I think it could be aligned to one common method.~~

### DIFF
--- a/templates/etc/systemd/system/docker-run.erb
+++ b/templates/etc/systemd/system/docker-run.erb
@@ -48,7 +48,11 @@ ExecStart=/usr/bin/<%= @docker_command %> start -a <%= @sanitised_title %>
 <% end -%>
 <%- if @before_stop %>ExecStop=-<%= @before_stop %>
 <% end -%>
+<<<<<<< HEAD
 ExecStop=-/usr/bin/<%= @docker_command %> stop --time=<%= @stop_wait_time %> <%= @sanitised_title %>
+=======
+ExecStop=-/usr/bin/<%= @docker_command %> stop <% if @stop_timeout != false %>-t <%= @stop_timeout %> <% end %><%= @sanitised_title %>
+>>>>>>> fe4c6e7... Added docker stop timeout possibility
 <%- if @remove_container_on_stop %>ExecStop=-/usr/bin/<%= @docker_command %> rm <% if @remove_volume_on_stop %>-v<% end %> <%= @sanitised_title %>
 <% end -%>
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,7 @@ with:
     bundle exec rake beaker
 
 This will run the tests on an Ubuntu 12.04 virtual machine. You can also
-run the integration tests against Centos 6.5 with:
+run the integration tests against Centos 7.0 with:
 
     BEAKER_set=centos-70-x64 bundle exec rake beaker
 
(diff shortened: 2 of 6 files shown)


templates/etc/systemd/system/docker-run.erb
```diff
@@ -39,6 +39,9 @@ ExecStart=/usr/bin/<%= @docker_command %> run \
         <% if @command %> <%= @command %><% end %>
 <%- if @before_stop %>ExecStop=-<%= @before_stop %>
 <% end -%>
+<% if @stop_before_kill -%>
+ExecStop=-/usr/bin/<%= @docker_command %> stop <% if @stop_timeout != false %>-t <%= @stop_timeout %> <% end %><%= @sanitised_title %>
+<%- end -%>
 ExecStop=-/usr/bin/<%= @docker_command %> kill <%= @sanitised_title %>
```
dimonzozo: Hi @lkosz, killing after stop is unnecessary, because `docker stop` will [kill container after timeout](https://docs.docker.com/engine/reference/commandline/stop/).

Configuring stop timeout seems reasonable for me, but I think it will be better to remove this line with `docker kill`.

lkosz: Yes, that line with docker kill should be removed. But we have to wait what garethr say because that change seems to be done intentionally. On the other hand when you look into initscript template, you'll notice that there is no docker kill...

maxd:
I think need to remove `docker kill` and leave `docker stop -t <timeout>` only and remove useless in this case `stop_before_kill` option. @garethr what do you think?

maxd:
@lkosz Could you please rebase from master this PR (this is required to re-run Travis CI)?

lkosz:
@maxd done

garethr:
@maxd it looks like this PR has pulled in some extra commits form elsewhere. If you could rebase against master and maybe squash the commits? That should make it easier to review. Thanks

maxd:
@garethr @lkosz is author if this PR :)

lkosz:
@garethr @maxd rebase done

maxd:
@lkosz Seems like this Travis CI problem related to Gentoo init.d script (it was suddenly added several days ago). Could you please add `-t <timeout>` option to [this init.d script](https://github.com/garethr/garethr-docker/blob/master/templates/etc/init.d/docker-run.gentoo.erb#L69) too? I have made these changes locally and all test has been passed successfully.

lkosz:
@maxd done :)

lkosz:
Yesterday I've rebased from master again and since that moment I have failing tests. Problem is that on two different machines tests passes.
@garethr @maxd do you have any idea what's wrong?

maxd:
@garethr @lkosz I don't understand why but Travis CI install different versions of `rspec`. Grep `rspec-core` in [valid](https://s3.amazonaws.com/archive.travis-ci.org/jobs/121363720/log.txt) and [invalid](https://s3.amazonaws.com/archive.travis-ci.org/jobs/122703907/log.txt) build logs.

lkosz:
@garethr @maxd tests passed. PR is ready for accept :)

lkosz:
Closing. Stop timeout functionality added in version 5.3

PR #449, opened 2016-02-26
Stop container in ExecStop instead of kill

dimonzozo:
`docker kill` sends SIGKILL to running container

`docker stop` sends SIGTERM signal to running container, waits for timeout (10 seconds default)
and then sends SIGKILL

Killing container right away seems not to be a good idea, because it can interrupt some cleanup process, writing to files, broke DBs and etc.

`stop` was changed to `kill` in c2ea3e5d2db6dde9fdec2138efe711df1490b471 and motivation was to mimic init.d behavior, but init.d is not doing kill, see [init.d#stop](https://github.com/garethr/garethr-docker/blob/master/templates/etc/init.d/docker-run.erb#L106).


dimonzozo:
Closing this in favor of #446.

PR #450, opened 2016-02-26
Run `metadata_lint` rake task instead `metadata`

dimonzozo:
https://github.com/puppetlabs/puppetlabs_spec_helper/commit/40546a44524a494142628010c6cab7dc9f24b991 commit removes metadata task and now we should run metadata_lint instead

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -50,7 +50,7 @@ begin
     :syntax,
     :lint,
     :parallel_spec,
-    :metadata,
+    :metadata_lint,
   ]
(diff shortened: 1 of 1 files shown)


garethr:
@dimonzozo thanks, I've fixed this in a slightly different (and backwards compatible) way in https://github.com/garethr/garethr-docker/commit/a59b297bae2d8a2e82b7698202e5d936589ec179. I basically consider that a bug in puppetlabs_spec_helper :(

PR #453, opened 2016-02-29
Don't quote dependency references

PierreR:
'language-puppet' chokes on this (rightly I would argue).

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -7,7 +7,7 @@
 #   Which tcp port, if any, to bind the docker service to.
 #
 # [*ip_forward*]
-#   This flag interacts with the IP forwarding setting on 
+#   This flag interacts with the IP forwarding setting on
 #   your host system's kernel
 #
 # [*iptables*]
@@ -120,9 +120,8 @@ class docker::service (
   }
 
   $_manage_service = $manage_service ? {
-    false   => undef,
-    true    => 'Service[docker]',
-    default => undef,
+    true    => Service['docker'],
+    default => [],
   }
 
   if $::osfamily == 'RedHat' {


garethr:
@PierreR ooh, interesting. Thanks. That might be a Puppet or a spec issue (as it appears to work correctly). I'll look to merge this once I get to testing it. Pinging @hlindberg in case it's of interest: should both of these work or just one?

PierreR:
I have also created an issue in `language-puppet` to clarify what this tool should do in this situation:

https://github.com/bartavelle/language-puppet/issues/176

PierreR:
By the way is the `default` case any helpful in
```
 $_manage_service = $manage_service ? {
    false   => undef,
    true    => 'Service[docker]',
    default => undef,
  }
```

hlindberg:
In Puppet 3.x most data types are represented with a String. This includes references like `File[foo]`.
We would like to deprecate the string form for references as it leads to unwanted (and crazy side effects).  While it works when assigning to a meta attribute like `require` it is just a string otherwise.

hlindberg:
The value switch is perhaps better written like this:
```
$_manage_service = $manage_service ? {
    true    => 'Service[docker]',
    default => undef,
  }
```
Or even simpler
```
$_manage_service = if $manage_service == true { Service[docker] }
```
The `if` produced the type reference or `undef` (an else part producing `undef` is not needed).

This assumes puppet 4.x language specification.

garethr:
Thanks @hlindberg. I'm currently supporting 3.8 as well unfortunately due to weight of numbers.

PierreR:
@hlindberg Thanks for the clarification. Very helpful indeed.

@garethr We are going to allow the reference string in the `permissive` mode of `language-puppet` in a future version. In the meanwhile, are you willing to accept the PR ? I have just changed the value switch as suggested by @hlindberg .

manifests/service.pp
```diff
@@ -120,8 +120,7 @@
   }
 
   $_manage_service = $manage_service ? {
-    false   => undef,
-    true    => 'Service[docker]',
+    true    => Service['docker'],
     default => undef,
```
jovandeginste: 'undef' should be '[]' (Otherwise, I get this error:

```
Error: Parameter notify failed on File[/etc/sysconfig/docker-storage-setup]: No title provided and :undef is not a valid resource reference at <xxx>docker/manifests/service.pp:135
```

PierreR: @jovandeginste  Thanks I have just forced a push with this change. Is it looking good ?

I am actually a bit surprised as the PR did not change anything in that regard. I mean it was `undef` before.

PierreR: @garethr Can you confirm it is best with `[]` ?

jovandeginste: Yes, it was. I was just testing the newer versions of the module (I fixed
my git sha) and had the exact same issue as you with the quoted resources.
Only when I fixed that part, I thought to try with no resources (manage =>
false) and got the next problem. So I thought I would add this comment to
your PR :-)

Met vriendelijke groet,
Jo Vandeginste

jo.vandeginste@gmail.com

2016-02-29 16:15 GMT+01:00 Pi3r <notifications@github.com>:

jfroche:
+1 for us as well

PR #456, opened 2016-03-06
not all systems have Internet access, this provides an opportunity fo…

findtopher:
minor refactor to allow override of docker-compose URL (production systems in our environment don't have outgoing Internet access, and as such we deploy from an internal web host).

I know that this is a relatively common practice, so this small refactor allows for using an internal URL for docker-compose installation.

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -19,7 +19,7 @@
 #
 class docker::compose(
   $ensure = 'present',
-  $version = $docker::params::compose_version
+  $version = $docker::params::compose_version,
   $install_from = "https://github.com/docker/compose/releases/download/${version}/docker-compose-${::kernel}-x86_64"
 ) inherits docker::params {
   validate_string($version)
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -13,9 +13,14 @@
 #   The version of Docker Compose to install.
 #   Defaults to the value set in $docker::params::compose_version
 #
+# [*install_from*]
+#   The URL to use to install docker-compose.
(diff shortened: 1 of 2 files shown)


manifests/compose.pp
```diff
@@ -13,9 +13,14 @@
 #   The version of Docker Compose to install.
 #   Defaults to the value set in $docker::params::compose_version
 #
+# [*install_from*]
```
garethr: `install_url` is probably a better name for this, as is describes that the param should be in the name, rather than relying on the description.

manifests/compose.pp
```diff
@@ -13,9 +13,14 @@
...
+#   The URL to use to install docker-compose.
+#   Defaults to the URL recommended by https://docs.docker.com/compose/install/
+#
 class docker::compose(
   $ensure = 'present',
-  $version = $docker::params::compose_version
+  $version = $docker::params::compose_version,
+  $install_from = "https://github.com/docker/compose/releases/download/${version}/docker-compose-${::kernel}-x86_64"
 ) inherits docker::params {
   validate_string($version)
```
garethr: If you could add a quick `validate_string` for the new `install_url` property as well.

garethr:
@findtopher thanks for this, appreciated. I've left a few quick comments. It would also be useful to add some simple unit tests to the test suite. These really help with keeping things working with all the changes being made by different folks to the module. If you need some pointers with the testing just let me know.

findtopher:
@garethr - unit tests included (along with a painful learning cycle about using parameters in other parameters).  This should be ready now; please let me know if you see testing gaps you'd like me to address.

PR #458, opened 2016-03-07
Added docker_network wrapper.

furhouse:
Included a docker_network type wrapper for hiera.
Also improved the README regarding hiera usage because:

```
docker::images:
  ubuntu:
    image_tag: 'precise'
```

produces an error: `hiera::images: produces Error: Must pass images to Class[Docker::Images]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,14 +368,14 @@ You can do that on the `docker` class like so:
 extra_parameters => '--cluster-store=<backend>://172.17.8.101:<port> --cluster-advertise=<interface>:2376'
 ```
 
-If using hiera, there's a `docker::network` class you can configure, for example:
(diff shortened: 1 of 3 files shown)


garethr:
Thanks. Appreciated. I noticed the `images` helper is plural, but the network one is singular. It would be useful to match that up, so `docker::networks` rather than `docker::network`. If you could make that quick change I'll get this merged in. Thanks in particular for the documentation updates too.

furhouse:
It's rewritten in plural, didn't notice it myself either. You're welcome!

furhouse:
Excuse me for the dirty commit history.

garethr:
@furhouse no problem. Thanks again for the work on adding this.

Issue #455, opened 2016-03-01
puppetlabs/apt 1.8.0 only


damiandennis:
Currently this module is only compatible with puppetlabs/apt v1.8.0 and no later. As you can see in the changelog 
https://github.com/puppetlabs/puppetlabs-apt/blob/master/CHANGELOG.md
they have removed a bunch of parameters in 2.0.0

JPalmerGithub:
Hi Damian & Gareth,
Yes, I have experienced similar issues. However, managed to fix it as the diff below.
I am sure it can be more elaborated, as parameterising etc.

garethr-docker (v5.1.0)
puppetlabs-apt (v2.2.2)
PUPPET Enterprise v2015.3.1
Ubuntu 14.04.4 LTS

[docker.txt](https://github.com/garethr/garethr-docker/files/159059/docker.txt)

garethr:
Closing in favour of #459. The issue is a backwards incompatible release of the apt module is coming at some point.

PR #459, opened 2016-03-08
Removing deprecated puppetlabs-apt properties to remove deprecation notices

tizzo:
The `key_source` and `required_packages` properties are deprecated and `key` should now be a hash instead of a string. I'm not sure the best way to handle the required packages so I've moved them to an `ensure_packages` call for now and made that a requirement of the apt resource.

Addresses #455.

### DIFF
--- a/manifests/repos.pp
+++ b/manifests/repos.pp
@@ -27,15 +27,15 @@ class docker::repos {
       if ($docker::use_upstream_package_source) {
         ensure_packages(['debian-keyring', 'debian-archive-keyring'])
         apt::source { 'docker':
-          location          => $location,
-          release           => $docker::package_release,
-          repos             => $docker::package_repos,
-          key               => {
+          location => $location,
+          release  => $docker::package_release,
+          repos    => $docker::package_repos,
+          key      => {
             'id'     => $package_key,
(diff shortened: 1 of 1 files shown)


garethr:
Also discussed in #315

manifests/repos.pp
```diff
@@ -25,15 +25,20 @@
...
-          include_src       => false,
+          location => $location,
+          release  => $docker::package_release,
+          repos    => $docker::package_repos,
+          key      => {
+            'id'     => $package_key,
+            'server' => 'hkp://keyserver.ubuntu.com:80',
+          },
+          pin      => '10',
+          require  => [
```
fessyfoo: the `requires` here exacerbates the issue discussed in #369  

also note https://github.com/puppetlabs/puppetlabs-apt/pull/427#issuecomment-75166159



> The idea here was that for some repositories you need a package installed that provides the keys for the repo. In the  case of the main Debian repositories, the debian(-archive)-keyring packages.
>
However, if you manage to install a Debian system without those packages you deserve what's coming to you, a world of pain. 

I suggest not managing those packages in this puppet-module.

fuzzycheck:
Is it going to be merged?

template-provider:
+1 ... can this be merged

PR #460, opened 2016-03-09
Updates `--cpuset` to `--cpuset-cpus`

petems:
- --cpuset was deprecated in 1.8.0, removed in 1.10.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,8 @@ docker::run { 'helloworld':
   volumes         => ['/var/lib/couchdb', '/var/log'],
   volumes_from    => '6446ea52fbc9',
   memory_limit    => '10m', # (format: '<number><unit>', where unit = b, k, m or g)
-  cpuset          => ['0', '3'],
+  cpuset          => ['0', '3'], # Docker < 1.10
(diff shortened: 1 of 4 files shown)


garethr:
@petems could you change this so as to change the implementation without changing the (puppet) interface?

I don't see a good reason to change the puppet interface as the change there is backwards incompatible. But we can abstract users from that.

If you _really_ want you could add `cpuset_cpus` as an additional param, have the code pick one and fail if the user sets both. But I'm not too bothered about that personally.

petems:
@garethr Fixed! :smile:

garethr:
mmm, this forces the burden of caring about this change in the docker CLI interface onto the user, which I don't like.

Personally I don't think this needs the new top level param, but if you're adding that it should simply be an alias for the existing param, and error if you provide both. The underlying implementation can chance to the new flags, I'm happy not supporting before 1.8.

petems:
Could possibly cause confusion though, as `--cpuset-mems` also exists now?

rogierslag:
Any update? The feature is unusable for all current docker versions...

petems:
Closing out old PR's, this is many years after the fact so I'm assuming it's super out of date! Closing 👍

PR #461, opened 2016-03-10
Docker CS source location

wazoo:
The docker CS source version was out of date so I updated that and added the ability to override it so I don't have to create a pull request every time Docker CS is updated.  I also added the new params to the docs.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -38,7 +38,7 @@
 #   The default is true.
 #
 # [*bridge*]
-#   Attach containers to a pre-existing network bridge 
+#   Attach containers to a pre-existing network bridge
 #   use 'none' to disable container networking
 #   Defaults to undefined.
 #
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -110,9 +110,15 @@
       $repo_opt = undef
       $nowarn_kernel = false
 
-      $package_cs_source_location = 'https://packages.docker.com/1.9/apt/repo'
-      $package_cs_key_source = 'https://packages.docker.com/1.9/apt/gpg'
-      $package_cs_key = '0xee6d536cf7dc86e2d7d56f59a178ac6c6238f52e'
+      unless $package_cs_source_location {
```
garethr: You shouldn't need the unless here. These set the defaults, which are then overridden if required in the init.pp.

manifests/init.pp
```diff
@@ -122,6 +122,16 @@
 #   If you're using an upstream package source, what is it's
 #   location. Defaults to https://get.docker.com/ubuntu on Debian
 #
+# [*package_cs_source_location*]
+#   If you want to override the hard coded Docer CS repo (for a newer version or
```
garethr: Typo, `Docer` should be `Docker`

garethr:
Hi @wazoo. Much appreciated. This currently fails the tests in Travis unfortunately. It would also be useful to add a unit test to verify this functionality and protect it from being broken in the future. I'm happy to help if you need a few pointers. Thanks again.

wazoo:
Thanks for the feedback, I will make those changes and update this when thats done.

PR #468, opened 2016-03-11
TLS

luckyraul:
@garethr I am not confident for supporting all OS, but it should be fine

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -450,7 +450,9 @@ class docker(
   }
 
   if($tls_enable) {
-    validate_array($tcp_bind)
+    if(!$tcp_bind) {
+        fail('You need to provide tcp bind parameter for TLS.')
+    }
     validate_string($tls_cacert)
     validate_string($tls_cert)
     validate_string($tls_key)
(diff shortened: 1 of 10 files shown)


JWGmeligMeyling:
Hi @luckyraul, the solution looks promising to me. I think this PR can be improved by adding a test and information to the README on how to use / generate these certificates.

manifests/init.pp
```diff
@@ -423,6 +449,13 @@
     fail('Values for dm_ variables will be ignored unless storage_driver is set to devicemapper.')
   }
 
+  if($tls_enable) {
```
garethr: It would also be nice to warn if `tls_*` are set but `tls_enable` is set to false, as they then get silently ignored.

garethr: Actually, these have sane defaults so this is probably fine.

garethr:
This looks great. Thanks @luckyraul. It would be really useful to have some simple unit tests added to the existing test suite to verify this, and help make sure it doesn't get accidentally broken. The CONTRIBUTING doc has some details but please do ask any questions if you're not too familiar with Puppet testing.

JWGmeligMeyling:
@luckyraul Thanks for adding the tests, and the build passes now as well. @garethr Are the unit tests sufficient or should Beaker integration tests also be added to ensure Docker indeed uses TLS now? Or will we test this manually?

garethr:
@luckyraul this looks great. Thanks for adding the tests as well.

PR #469, opened 2016-03-13
add build-args property to docker::image

macwadu:


manifests/image.pp
```diff
@@ -74,10 +75,17 @@
...
+  if $build_args {
+    validate_array($build_args)
+    $build_args_string = join ( regsubst ($build_args, '^(.*)$', '--build-arg \1 '), '')
+  } else {
+    $build_args_string = ''
+  }
+
   if $docker_dir {
-    $image_install = "${docker_command} build -t ${image_arg} ${docker_dir}"
+    $image_install = "${docker_command} build ${build_args_string}-t ${image_arg} ${docker_dir}"
```
garethr: Looks like you might be missing a space before `-t`?

spec/defines/image_spec.rb
```diff
@@ -68,6 +68,25 @@
...
+    it { should contain_exec('docker build --build-arg FOO=bar --build-arg BAR=foo -t base - < Dockerfile') }
+  end
+
+  context 'with docker_dir => /tmp/docker_images/test1 and build_args => [FOO=bar, BAR=foo]' do
+    let(:params) { { 'docker_dir' => '/tmp/docker_images/test1', 'build_args' => ['FOO=bar', 'BAR=foo'] } }
+    it { should contain_exec('docker build --build-arg FOO=bar --build-arg BAR=foo -t base /tmp/docker_images/test1') }
+  end
+
+  context 'with docker_dir => /tmp/docker_images/test1 and build_args => FOO=bar' do
+    let(:params) { { 'docker_dir' => '/tmp/docker_images/test1', 'build_args' => 'FOO=bar' }}
```
garethr: You can support this formulation as well using the `any2array` function.

manifests/image.pp
macwadu: I'm dealing with the space when building the 'build_args_string', i don't want to put an extra space when build_args_string is empty. Do have a better way of dealing with that?

Still i saw the Ruby 1.8.7 build failing because of the rake version is that something i should try to fix on the Gemfile?

garethr: Ah, no worries. I'm just fixing the rake issue, rake 11.x shipped this weekend. You should be able to rebase with master once I get that fixed.

macwadu: Was something like this b2c7786538218e0d16e4a0a31cf5cbc44c952ea9 what you mean?

garethr: I _think_ (off the top of my head) you might not need the `!empty($build_args)` check or the second clause. Worth a try.

macwadu: I think both aren't needed but i may be wrong but this way it will always go into the if :( ). Let's wait for the tests on multiple version. Please feel free to point anything that you don't agree.

PR #471, opened 2016-03-13
Add support for docker's --bip and --mtu options

bpineau:
This may be needed to support software defined networks like flanneld,
or other custom network setups.


garethr:
@bpineau thanks, much appreciated.

PR #472, opened 2016-03-15
Give the possibility to override create_user

jfroche:
`create_user` is currently with default true and there was no way to
override this behaviour.

With this change we give the possibility to override the `create_user` value using
puppet automatic parameter lookup on the docker::config class.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -1,5 +1,9 @@
 # == Class: docker::config
 #
-class docker::config {
-  docker::system_user { $docker::docker_users: }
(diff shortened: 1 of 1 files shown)


PierreR:
:+1:

garethr:
@jfroche Thanks for this. Ideally this would be able to be controlled from the top level `docker` class. And have the default specified in the `params`. It would also be useful to have unit tests that exercise this functionality. You should be able to get started from the details in the CONTRIBUTING doc, but any questions do please let me know.

PR #473, opened 2016-03-15
Missing params variables

jfroche:
Some variables are used in both init.pp and params.pp but are not initialized correctly.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -24,6 +24,7 @@ class docker::params {
   $service_state                     = running
   $service_enable                    = true
   $service_provider                  = undef
+  $service_overrides_template        = undef
   $manage_service                    = true
   $root_dir                          = undef
   $tmp_dir                           = '/tmp/'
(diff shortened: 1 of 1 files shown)


garethr:
These are set on a per OS basis (ie. in the params file but further down). And as variables in Puppet are immutable, once set they can't be overridden. What problem do you see here? Are you seeing an error because these are missing?

jfroche:
In RedHat based version 6 the variable is used by service.pp and init.pp but is never defined in params.pp.

Puppet enables variable lookup ($docker::params::service_overrides_template) even on variable that doesn't exists. We would prefer to define this variable in any case (even as undef). This way we can't prevent unknown variable to be silently accepted.
We are using https://github.com/bartavelle/language-puppet to verify our puppet catalogs and this one is stricter on undefined variables.

jfroche:
See what you mean with immutable parameters in params.pp. We will fix these variables in redhat el6 case in another pull request.

PR #474, opened 2016-03-15
Fix missing variables on RedHat < 7.0

jfroche:
`service_provider` and `service_overrides_template` variables were not defined when used
on RedHat 6 even tough they are used in init.pp or service.pp


PierreR:
:+1:

PR #476, opened 2016-03-17
Ensure that containers are restart after the docker restarts

houtmanj:
If the module decides to reload the docker daemon then all containers are stopped,
this change ensures that all containers are started after the refresh of the service.

The functionality can be disabled with the restart_service_on_docker_refresh parameter in docker::run.
For it to work the docker_service parameter needs to be set, indicating that the docker daemon is managed by this module.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -26,6 +26,11 @@
 #   (optional) Whether or not to restart the service if the the generated init
 #   script changes.  Default: true
 #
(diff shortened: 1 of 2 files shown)


houtmanj:
Continuation of https://github.com/garethr/garethr-docker/pull/356

garethr:
@houtmanj thanks

PR #477, opened 2016-03-17
Configurable Service templates for docker::run

houtmanj:
Allow the templates for starting an docker container to be replaced.
The functionality is handy for adding additional logic to the templates
without complicating the module itself.

We currently use it for integration with consul.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -43,6 +43,10 @@
 # command. Useful for adding additional new or experimental options that the
 # module does not yet support.
 #
+# [*service_tempalte_override*]
+#   Specify an template that overrides the template used for systemd or init.d
+#   Default does not override the templates provided by the module.
+#
 define docker::run(
   $image,
   $ensure = 'present',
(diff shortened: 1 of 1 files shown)


houtmanj:
Continuation of https://github.com/garethr/garethr-docker/pull/355

manifests/run.pp
```diff
@@ -43,6 +43,10 @@
 # command. Useful for adding additional new or experimental options that the
 # module does not yet support.
 #
+# [*service_tempalte_override*]
```
garethr: spelling mistake, `tempalte` instead of `template`.

manifests/run.pp
```diff
@@ -43,6 +43,10 @@
...
+#   Specify an template that overrides the template used for systemd or init.d
```
garethr: grammar, should be `a template` rather than `an`

garethr: I also think it's worth being explicit this is the template path.

garethr:
@houtmanj thanks, this looks useful. I've left a few minor comments on the docs. If you could also add a simple unit test to demonstrate the functionality. This really helps with maintaining all the various code paths.

houtmanj:
This is my attempt at an test,  I wanted to check the contents on the file on the system. 
But I need some help to get that working.

garethr:
@houtmanj sorry for the long wait on getting back to the issues here. With regards a test I'd recommend starting with just adding an rspec-puppet unit test rather than the more involved acceptance tests. See the examples in `spec/defines/run_spec.rb`.

PR #480, opened 2016-03-21
Don't interfere with apt sources I might configure

benlangfeld:
This dependency only applies to sources managed by this module, so set them explicitly.


benlangfeld:
Closing in favour of https://github.com/garethr/garethr-docker/pull/482

PR #481, opened 2016-03-21
Support --bip and --mtu on Ubuntu

benlangfeld:
This wasn't included in https://github.com/garethr/garethr-docker/commit/9e0ffcc12982dfef23724c196e70965e25a1962d


garethr:
@benlangfeld thanks for catching. Appreciated.

PR #488, opened 2016-03-26
fix #447 by using docker create on prestart, and docker start instead…

### LINKED ISSUE #447: Unable to start container with "remove_container_on_start = false"
See [this](https://github.com/garethr/garethr-docker/issues/422#issuecomment-181587473) comment. It doesn't look like docker::run is working with `remove_container_on_start = false`... it fails for me on every reboot/provision assuming this option is set.

This is compounded by the fact that it seems docker::run almost requires use of the init scripts. The only way I can seem to bypass this is by setting `restart` option to a value, but then the same issues - docker::run will fail because of duplicate container names.

I feel like docker::run should be able to resolve this silently, and use something like `docker start` when an existing container is found. Thoughts?

wscheele:
… of docker run for start.

seems to work ok. i am not so happy with the create ExecStartPre in systemctl status that is Failed by default (on success). But the same approach is taken for the kill ExecStartPre.

### DIFF
--- a/templates/etc/init.d/docker-run.erb
+++ b/templates/etc/init.d/docker-run.erb
@@ -79,15 +79,25 @@ start() {
     <% end -%>
     <% if @remove_container_on_stop -%>
         $docker rm <% if @remove_volume_on_start %>-v<% end %> <%= @sanitised_title %> >/dev/null 2>&1
+    <% else %>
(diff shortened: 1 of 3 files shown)


garethr:
@wscheele thanks, much appreciated.

PR #495, opened 2016-04-07
Only include debian backports if $docker::use_upstream_package_source is true

davewongillies:
When I set `$docker::use_upstream_package_source` to `false`, I don't expect this module to still attempt to manage an apt repo for me, even if it is backports.

I've moved the inclusion of backports down so that its only included if `$docker::use_upstream_package_source` is `true`.


davewongillies:
Nevermind

PR #496, opened 2016-04-07
Give --tls option to docker daemon when tls_enable => true

hanazuki:
Under the conditions `tls_enable => true, tls_verify => false`, docker daemon listens TCP socket without TLS enabled.
Explicitly giving `--tls` option, this patch makes the daemon to listen with TLS even when the `--tlsverify` flag is not set (Note that `--tlsverify` implies `--tls`).


garethr:
@hanazuki thanks, good spot. Thanks in particular for adding the tests. Much appreciated.

PR #498, opened 2016-04-12
Systemd Service Overrides for Ubuntu

cablespaghetti:
The system service overrides file was only used on Debian versions with systemd not Ubuntu. This causes none of the options in /etc/default/docker to work when systemd is used to start the service.

Thanks
Sam


garethr:
Thanks @cablespaghetti

PR #499, opened 2016-04-13
Fix some undefined variables

joshuaspence:
When running `puppet` with the optional `strict_variables` mode enabled, there are failures due to variables being undefined:

```
Error: Evaluation Error: Unknown variable: 'docker::params::service_config'.
Error: Evaluation Error: Unknown variable: 'docker::params::service_overrides_template'
```

Ensure that these variables are defined, regardless of the `$::osfamily`.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -91,6 +91,7 @@ class docker::params {
             $service_hasstatus       = true
             $service_hasrestart      = false
           }
+          $service_overrides_template = undef
         }
         default: {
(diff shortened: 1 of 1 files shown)


mcanevet:
@garethr could you please merge this? Currently your module does not work when `strict_variables` is enabled. (I don't know why the unit tests does not show it though...)

garethr:
@joshuaspence thanks for those fixes. I'm hoping to refactor the test suite some and make sure it catches those issues in the future too.

PR #500, opened 2016-04-17
Redirect inspect errors to /dev/null for status checks

adepretis:
This avoids `Error: No such image or container` when starting or restarting


garethr:
Thanks @adepretis

PR #504, opened 2016-04-20
fixed: refresh on docker::image should rebuild the image

danielpanteleit:
Refrehes on docker::image wouldn't rebuild the image. The culprit is the `unless` parameter in the install `exec`:

```
    exec { $image_install:
      unless      => $image_find,
...
```

My solution is not that elegant, but should work: I added a clone of the `exec` which will only run on refreshes and only if the image is found.

Another solution would be to add an `exec` for the image_find which would run every time and refresh the image_install if no image is found. Something like this:

```
exec { $image_find:
  command => 'true',
  unless => $image_find,
  notify => Exec[$image_install],
}
```

That way the original exec could change to `refreshonly` without the `unless`.

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -107,6 +107,17 @@ define docker::image(
       returns     => ['0', '1'],
       require     => File['/usr/local/bin/update_docker_image.sh'],
     }
+
+    exec { "refresh - ${image_install}":
+      command     => $image_install,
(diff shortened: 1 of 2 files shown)


gdippolito:
Is there any ETA to have this request merged? 

We are having the same issue and it would be nice to have this fixed.

PR #510, opened 2016-05-03
Fix docker run when using docker swarm

heliocentric:
docker only accepts -H before the subcommand on recent versions.
In order to fix this I've removed socket_connect from the run flags and added
it to the docker_command variable.

This has the added benefit of prefixing every command for docker
run in the init scripts, with the socket_connect option. This can
prevent weird situations where docker stop/kill will run against a container
the local docker swarm node doesn't know about.


garethr:
@heliocentric thanks, nice fix. Much appreciated.

PR #512, opened 2016-05-04
Re-enabled lint check case_without_default 

ttarczynski:
Small change to obey this style guide requirement:
- https://docs.puppet.com/guides/style_guide.html#defaults-for-case-statements-and-selectors
- http://puppet-lint.com/checks/case_without_default/


ttarczynski:
@garethr can we have this small fix merged?

garethr:
Thanks @ttarczynski

Issue #514, opened 2016-05-17
docker_compose doesn't support v2 compose file syntax


zsolt-keseru-epam:
After a quick look I suppose the following issue exists (fixme if I'm wrong), but haven't actually tried :

https://github.com/garethr/garethr-docker/blob/master/lib/puppet/provider/docker_compose/ruby.rb#L18
Here in the 'exists?' check tries to iterate over top-level keys of the yaml (v1 syntax), since v2 container definitions moved under 'services' key and a 'version: "2"' top-level key-value has been introduced to distinguish between versions.

v1 syntax will be deprecated, the following page mentions "supported up to 1.6.x", although it still works in 1.7.1: https://docs.docker.com/compose/compose-file/#version-1

nparley:
@garethr Would be great if this could be looked at or merged. I can confirm version 2 syntax compose files do not play nice at the moment and I have to use version 2 for the docker networking. 

```
Info: Checking for compose service version
Info: Checking for compose service services
Info: Checking for compose service networks
```

garethr:
I've merged #517 into master so master should now work correctly with v2 compose files. I'll look to get a new release out soon.

nparley:
Thanks Gareth.

PR #517, opened 2016-05-18
Fix compose file v2 syntax compatibility

zsolt-keseru-epam:
#514


garethr:
Thanks @zsolt-keseru-epam. And thanks for adding the tests as well, much appreciated.

PR #521, opened 2016-05-27
Updating doc to reflect possibilities with 'net'

tux-o-matic:
The docker::run resource exposes a 'net' parameter but the README only mentions extra_parameters to set it. Plus the block regarding docker::networks doesn't mention at to reference a custom network from an instance.


garethr:
Thanks for the update. The original examples in the README could definitely do with an update. Thanks again.

PR #523, opened 2016-06-01
Added support for splunk to log driver

eperdeme:
Due to input validation you can not use the splunk or any new log drivers.
Rather than rip it out as I guess it was wanted, I have added the new splunk log driver in to the validator.


garethr:
@eperdeme thanks, appreciated.

PR #524, opened 2016-06-07
Add Docker Daemon's ICC parameter

lazyfrosch:
To control unrestricted inter-container and Docker daemon host communication (with iptables).

Default is true, but we will only set it if specified by user.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -57,6 +57,11 @@
 #   Enable IP masquerading for bridge's IP range.
 #   The default is true.
 #
+# [*icc*]
+#   Enable or disable Docker's unrestricted inter-container and Docker daemon host communication.
+#   (Requires iptables=true to disable)
+#   Default is undef. (Docker daemon's default is true)
+#
 # [*bip*]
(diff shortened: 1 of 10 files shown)


manifests/init.pp
```diff
@@ -57,6 +57,11 @@
 #   Enable IP masquerading for bridge's IP range.
 #   The default is true.
 #
+# [*icc*]
+#   Enable Docker's unrestricted inter-container and Docker daemon host communication.
+#   (Requires iptables=true to disable)
+#   Default is true.
```
garethr: This says the default is true, but in params it's set to undefined. I think it's worth being clear, explicit true is different to undefined and implicitly true because of defaults in Docker for instance.

garethr:
@lazyfrosch thanks, much appreciated. One small comment if you could resolve or let me know and I'll get this merged. Cheers

lazyfrosch:
I clarified the default setting (what Docker's default is)

Technically the module should **not** set ip{tables,_masq,_forward} by default also, but I think we shouldn't change that simply.

garethr:
Thanks @lazyfrosch

PR #527, opened 2016-06-08
Use the registry resource title for auth

ssube:
I have a use case where two different users need to register with the same registry. The old server-based naming of the auth resource wouldn't allow that, since the inner auth-exec had a duplicate title.

This changes the exec's title to be based on the surrounding resource's title, allowing manifests to make them unique if they want (otherwise it just defaults back to the server name).

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -54,7 +54,7 @@ define docker::registry(
       $auth_cmd = "${docker_command} logout ${server}"
   }
 
-  exec { "auth against ${server}":
+  exec { "${title} auth":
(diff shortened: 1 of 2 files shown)


mterzo:
I had the same problem with this as well. I put the local_user in the exec. Was about to submit a patch.

ssube:
I considered putting the user into the `exec`, but thought that using the parent resource's title would be better for guaranteeing uniqueness and be more flexible if something unusual came up.

garethr:
@ssube neat solution, thanks. Appreciated.

PR #530, opened 2016-06-11
Ensure that curl is set up before calling the command

### LINKED ISSUE #529: The docker module relies on curl (for docker-compose) but doesn't ensure that curl is available
The docker::compose module should have `ensure_packages(['curl'])` to ensure that curl is available before running the curl command, as well as a `require => Package['curl']` on the exec itself.

ColinHebert:
Fixes #529


garethr:
@ColinHebert thanks for the catch and fix. Appreciated.

PR #531, opened 2016-06-13
Fix: exit 3 when image is pulled for the first time

Morsicus:
When i pull an image for the first time, exit code is 3. Now exit code is 1.


Morsicus:
.Config.Images contains parent's hash. Alpine/busybox have no parents. I think it's better to use .Id

garethr:
@Morsicus thanks for the fix!

PR #532, opened 2016-06-14
add --time=<> to docker stop

coeotic:
There is a use case in our environments, where we need the dockerized app to accept SIGINT (2), so it has a chance to clean up any side-effects in it's environment. This change provides time (default: 0 seconds) for the container to shut itself down.

Note: could not find testing for docker-run*.erb (e.g. $sanitised_title), so no new testing added.


garethr:
Thanks @coeotic, nice addition. Appreciated.

PR #536, opened 2016-06-20
add up_args to docker-compose up cmd

AblionGE:
Add an ```args``` option to be able to give options to the ```docker-compose``` command (especially ```--remove-orphans``` to the ```docker-compose up``` command).


garethr:
@AblionGE thanks, appreciated.

PR #539, opened 2016-06-20
Add configurable docker-storage-setup file path

### LINKED ISSUE #538: docker-storage-setup file needs to be configurable
RHEL offers a `docker-latest` package that currently install Docker 1.10. This package installs and expects an `/etc/sysconfig/docker-latest-storage-setup` file, but the module has the location `/etc/sysconfig/docker-storage-setup` set and non-configurable.

walkamongus:
This fixes #538 and allows a user to specify the path for the
docker-storage-setup file.


garethr:
Thanks @walkamongus, this looks good to me. If you could just rebase against master as it clashes with some just-merged changes. Any problems let me know.

PR #547, opened 2016-07-09
Fix lint and rspec test failures

dvorak:
This fixes a bunch of warnings about long (>140 character) lines and
also fully qualifies some class references that was causing spurious
failures.


dvorak:
This fixes #548

garethr:
Thanks @dvorak

PR #550, opened 2016-07-12
Add refresh to docker compose

nparley:
I wanted docker compose to rebuild when notified by a vcsrepo change so added this refresh method. Now:

```puppet
vcsrepo { 'repo': 
  ensure   => latest,
  ...
} ~>
docker_compose { '/repo/docker-compose.yml':
  ensure  => present,
}
```

will cause the docker compose file to be stopped, rebuilt and restarted when the repo updates. This could also be used to rebuild compose if the docker Files were being managed by docker by notifying from file change.


garethr:
@nparley thanks, nice addition.

PR #552, opened 2016-07-14
change http to https in README

rchicoli:
this should fix some silly warnings someone gets when trying to access the ```README.md``` in a private git server: ```Parts of this Page are not secure```


garethr:
Thanks @rchicoli

PR #555, opened 2016-07-18
Load pry optionally

DavidS:
pry is a development gem, therefore not available during acceptance testing.
Keep it in with an optional require to keep it around when debugging locally.


DavidS:
[now with local verification that it can run beaker]

PR #557, opened 2016-07-19
Allow usage of proxy for Docker Compose download

mikey179:
This PR enables to use a proxy for downloading Docker Compose. Sometimes the server in which the Puppet code runs is behind a proxy and can't access the download target directly, but it is not required to specify a proxy for the whole system. In such cases it can be helpful to specify the proxy for this download only.

Currently the tests with _PUPPET_GEM_VERSION="~> 3.0" STRICT_VARIABLES=yes_ fail (see https://travis-ci.org/mikey179/garethr-docker/builds/145836854) - I'm a bit unsure what to do about it. From what I see I'd need to specify all variables from _manifests/params.pp_ in the `let(:facts)` section, but is this really the way it should be done?

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -17,22 +17,35 @@
 #   The path where to install Docker Compose.
 #   Defaults to the value set in $docker::params::compose_install_path
 #
+# [*proxy*]
+#   Proxy to use for downloading Docker Compose.
+#
 class docker::compose(
   $ensure = 'present',
   $version = $docker::params::compose_version,
-  $install_path = $docker::params::compose_install_path
+  $install_path = $docker::params::compose_install_path,
+  $proxy = undef
 ) inherits docker::params {
   validate_string($version)
   validate_re($ensure, '^(present|absent)$')
   validate_absolute_path($install_path)
+  if $proxy != undef {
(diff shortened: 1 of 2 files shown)


garethr:
Hi @mikey179. First up, thanks for submitting and apologies for taking so long to get back to you on the PR. This looks great, thanks in particular for adding the tests.

As you noted you hit an issue under strict variables with the tests. I need to port the tests to use `rspec-puppet-facts` which will fix this, but in the meantime if you just use the following in your test case that would be fine.

```puppet
let(:facts) { {
  :osfamily               => 'Debian',
  :operatingsystem        => 'Ubuntu',
  :lsbdistid              => 'Ubuntu',
  :lsbdistcodename        => 'maverick',
  :kernelrelease          => '3.8.0-29-generic',
  :operatingsystemrelease => '10.04',
  :operatingsystemmajrelease => '10',
} }
```

Thanks again, it's much appreciated.

mikey179:
Hi @garethr, thanks for the feedback. I added the facts you suggested, so tests are green now. Please let me know if there's anything else to be done before it can be merged.

garethr:
@mikey179 thanks, much appreciated

PR #560, opened 2016-07-20
fix #559: add param for restarting containers

### LINKED ISSUE #559: Docker-Compose refresh: do not restart all containers
As pointed in issue #553 , it can be useful to not restart all containers when a refresh event occurs on the ```Docker_Compose``` command.

A parameter can fix the problem. Any other idea ?

AblionGE:
Add a parameter to select if a refresh event must restart all containers or simply run the `docker-compose up` command.

### DIFF
--- a/lib/puppet/type/docker_compose.rb
+++ b/lib/puppet/type/docker_compose.rb
@@ -4,7 +4,7 @@ Puppet::Type.newtype(:docker_compose) do
   ensurable
 
   def refresh
-      provider.restart
+    provider.restart
   end
 
   newparam(:name) do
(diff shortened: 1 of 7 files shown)


lib/puppet/type/docker_compose.rb
```diff
@@ -38,6 +38,13 @@ def refresh
 		end
 	end
 
+	newparam(:restart) do
```
garethr: I think this should have a default value of true. This would make it optional and preserve backwards compatibility I think?

lib/puppet/type/docker_compose.rb
```diff
@@ -38,6 +38,13 @@ def refresh
...
+    desc 'Argument to set if containers have to be killed and restarted on refresh event'
```
garethr, line 42: Looks like this and the next few lines have white space issues?

garethr (changes requested):
Thanks, this looks like a useful addition. Apologies for taking so long to get around to reviewing it.

I've left a couple of minor comments, mainly concerning making it backwards compatible and providing a default value for the new param. Any questions at all please let me know. And thanks again.

lib/puppet/type/docker_compose.rb
AblionGE: @garethr Thanks for your comment. Is [this line](https://github.com/AblionGE/garethr-docker/blob/7169f04ab27f0064362a1e9a10db9972c85fd651/lib/puppet/provider/docker_compose/ruby.rb#L66) enough for default value ?

lib/puppet/type/docker_compose.rb
AblionGE: Yes! Fixed! Thanks!

AblionGE:
Squashed to be merged.

PR #561, opened 2016-07-26
add initial support for sles 12.0

arki91:
Hi,

I added support for Sles 12.0 running with systemd. 
The module is tested (installation, image pull and docker run) and is currently used in our organization.

Docker-compose is currently untested, because the default version in sles 12 is atm broken. 
Networks is untested, we used defaults.
Registry is untested, but shouldn't pose any problems, as it is os unspecific.
Repos is not used, as docker is already in Sles default repositories and is activated by default (our installation, not 100% sure, might need to add support later).
system_user is untested.

Please inform me if you think something needs improvement.

Kind regards, 
Alex

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -415,8 +415,8 @@ class docker(
 ) inherits docker::params {
 
   validate_string($version)
-  validate_re($::osfamily, '^(Debian|RedHat|Archlinux|Gentoo)$',
-              'This module only works on Debian or Red Hat based systems or on Archlinux as on Gentoo.')
+  validate_re($::osfamily, '^(Debian|RedHat|Archlinux|Gentoo|Suse)$',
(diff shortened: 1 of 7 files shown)


garethr:
@arki91 first up, thanks for looking at this and apologies for taking so long to respond. This looks great. The only thing I'll need before merging is to update the test suite. This helps make sure things don't get broken by other contributions. This should just involve adding Suse to lines like https://github.com/garethr/garethr-docker/blob/master/spec/classes/docker_spec.rb#L5 to trigger the tests for Suse, and fixing up the strings in the tests https://github.com/garethr/garethr-docker/blob/master/spec/classes/docker_spec.rb#L1032. Thanks again, much appreciated. Any questions at all about the tests do let me know.

snahelou:
Hello

I tried to implement your tests https://github.com/garethr/garethr-docker/pull/635

Regards

Issue #564, opened 2016-07-30
docker::compose fails on systems which already have package curl managed via puppet.


rehanone:
docker::compose has an ensure_package call that fails on systems which already have package curl managed via puppet.

garethr:
Could you provide an example of the failure? That is exactly the issue `ensure_packages` is designed to avoid. https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/ensure_resource.rb#L14

mattbarrio:
@garethr this is caused by a module using the conventional package resource. You can duplicate this by doing:

`$packages = ['wget']

package { $packages:
  ensure => 'installed',
}

ensure_packages($packages)`

Unfortunately, the only real way for @rehanone to get around this is to upgrade/update whatever module is using the conventional package resource.

ryayon:
Hi,

Same issue here.
I cannot remove my `curl` entry as I install it on every machine.

jonathonmg:
@garethr I ran into this same issue because one of our system modules had this resource:

if ! defined(Package['curl']) { package { 'curl': ensure => latest, } }

If I change that line to this:
ensure_packages(['curl']) The resource collision error goes away. However if I call it like this:
ensure_packages(['curl']), {'ensure' => 'latest'}) Then I still get a duplicate resource error.
On further reading of the documentation you referenced in ensure_resource.rb this appears to be by design: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/ensure_resource.rb#L18

As such, IMO this kind of renders the function pretty useless and keeps, "if ! defined(Package..." as the more reliable pattern.

aaronbbrown:
This is also preventing us from upgrading to 5.3.0 due to the issue described in https://github.com/garethr/garethr-docker/issues/564#issuecomment-284849277.

PR #565, opened 2016-07-30
Added check to make sure if the package curl is already defined and p…

rehanone:
#564

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -27,7 +27,10 @@ class docker::compose(
   validate_absolute_path($install_path)
 
   if $ensure == 'present' {
-    ensure_packages(['curl'])
+
+    if ! defined(Package['curl']) {
+      ensure_packages(['curl'])
+    }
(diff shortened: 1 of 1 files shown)


freibuis:
isnt this what ensure_packages already does?
https://forge.puppet.com/puppetlabs/stdlib#ensure_packages

Thubo:
It happens whenever you have a statement such as

`package { curl: ensure => 'installed' }`

somewhere in your code.

If you replace the `package` statement by

`ensure_packages(['curl'])`

everything works as expected.

I guess this is (if it is a bug at all) an issue on the puppet or puppet-stdlib side.

The solution of @rehanone works as well - since it checks if the package was declared somewhere before. Idk for sure what is best practice in this case.

garethr:
Also see #617. If anyone is running into this issue if you could post a failing test case or gist of agent output and manifests so I can recreate that would be appreciated.

PR #570, opened 2016-08-05
Add ability to configure package pinning on Debian based systems.

gizmoguy:
An updated version of PR #86.

Allows us to reconfigure the apt pinning of the docker repo so that we can upgrade docker-engine.


garethr:
@gizmoguy thanks, this looks great. Much appreciated.

PR #571, opened 2016-08-08
Bump default docker-compose version to 1.8.0

Thubo:
See also https://github.com/docker/compose/releases


garethr:
Thanks for the heads up, and sorry for taking an age to get to the backlog of issues. I've bumped this to 1.9 which was recently released but thanks again for the PR.

PR #573, opened 2016-08-12
{WIP} Add support for Windows (10)

DLuCJ:
Current: 
    On Windows 10, docker-engine is supported using MobyLinuxVM and Hyper-V,
    as it can only run on Linux.  The VM is launched using the
    Docker for Windows app.  

    However, the app often fails on startup with the error: Failed to start VM “MobyLinuxVM”: 
    The VM couldn’t get an IP address after 60 tries

    Setup: VMWare Fusion 8
                Networking: Bridged autodetect
                Hypervisor applications enabled
                Hyper-V enabled

    The network adapter disable / enable toggle  before executing the Docker for Windows app
    does not always work.  I've found a few other little tricks that occasionally work.  A diff of the
    logs for passing and failing runs reports identical output until the passing run displays a hvsock 
    successful connection while the failing run is still waiting for an ip.  The Docker forums / github 
    issues do not have a commonly accepted fix. 

    For example: https://forums.docker.com/t/failed-to-start-vm-mobylinuxvm-the-vm-couldnt-get-an-ip-address-after-60-tries/19551

    Going to test this with a bare metal Win 10 install before digging in further.  
 
******************************************************************************************
General:
    * Docker_command has different value depending on platform
    * Windows paths separated from my machine
    * Facts added to facilitate the above two points
    * Spec tests added to validate the facts
    * First pass at cleaning up the Powershell code
    * A few Windows-only params have been set in params.pp

    * Prerequisite: Hyper-V enabled
    * Prerequisite: Powershell module installed
    * Prerequisite: 64bit Windows 10 Pro, Enterprise and
          Education (1511 November update, Build 10586 or later) 

Docker install:
    * Docker for Windows Stable MSI is downloaded from the Docker site    
    * Checksum    
    * MSI is installed through the Puppet default package provider    
          for Windows
    * After installation, Docker for Windows app is started.

Docker::image:
    * Pull tagged and untagged images  
    * Remove tagged and untagged images
    * If these images are already installed / removed, docker::image does nothing 
    * Docker::image can add images from docker file or docker dir 

Docker::run:
    * Run containers if restart is set (i.e no init script)
    * Checks for cidfile existence / cid match before executing

Docker Networks:
    * Specs pass, initial testing okay
    * No changes made to lib code

Docker Compose:
    * Specs pass, initial testing okay
    * Compose install / uninstall guarded on Windows, as this is managed together
          with the entire Docker package on Windows.
    * No changes made to lib code

Docker Private registry:
    * Attempting log in / log out of invalid registries gives expect output.
    * User param removed - cannot execute as other users on Windows
    
Docker::exec:
    * Can execute arbitrary commands in the context of a running container.

Todo:
    * Fix MobyLinuxVM ip address error
    * Cross-platform support (add back the non-Windows code paths that got removed)
    * Acceptance tests passing, add tests as needed, 
    * Double-check Windows params values in params.pp.
    * Add support for docker service / daemon configuration
    * Add support for docker containers managed by Windows SCM or maybe NSSM

    * Test Docker Private registry with a valid registry
    * Test Docker tar
    
Known Issues:
    * docker::image ensure latest does not validate 'unless' properly
    * Exec does not validate 'unless' properly


DLuCJ:
A few questions / notes here.

* If a tagged image is downloaded (ubuntu:precise for example), untagged pulls / removes 
   of that image do not work properly (ensure present / absent on ubuntu).  

   That is, docker::image will attempt to remove ubuntu (default:latest) even if ubuntu:latest 
   is not installed because the exec conditional matches the ubuntu in ubuntu:precise.  
   Likewise, docker::image will not install ubuntu (default:latest) even ubuntu:latest is 
   not installed. 

   Is this desired behavior?  Do we want to require that all image resources be tagged if there is 
   a tagged version of that image that is already downloaded?

* What is the Windows analogue to init.d/systemd/conf.d/etc service config files?
   This is needed for both container and daemon management.
   Can something like NSSM provide the same functionality as the config files?
   Can it do so at the container level?  
   Why not configure the daemon by writing daemon.json?

* Do we want to start lifting more code from manifests into lib type/provider?  If not, what's the 
   reasoning behind the split between networks / compose and the other types?  

* Both latest and beta MSIs are available for download.  Currently, the module is only 
   pulling stable.  Should we give the option to choose?

   How do you want to handle upgrades?  Currently, the user needs to ensure that the 
   downloaded MSI is absent in order to pull a new version.  They may also need to uninstall 
   the old version first.  I haven't tested this.   
   
* The issues with running defines/classes tests on my machine may just be related to env, as the
   defined types / docker class are resolved properly in Travis.  I also don't see any platform 
   guards against Windows that would result in this behavior.

* As far as I can tell, Travis is only running against puppet 3.8.7.  Do we want to test against 
   later versions as well? 

@garethr Any thoughts?

PR #576, opened 2016-08-16
Pull by image digest

RickyCook:
Allows pulling of images by digest so that you can be certain of image content

---

This change is [Reviewable](https://reviewable.io/reviews/garethr/garethr-docker/576)

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -14,6 +14,9 @@
 # [*image_tag*]
 #   If you want a specific tag of the image to be installed
 #
+# [*image_digest*]
+#   If you want a specific content digest of the image to be installed
+#
 # [*docker_file*]
 #   If you want to add a docker image from specific docker file
 #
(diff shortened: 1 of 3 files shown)


RickyCook:
Tests seem to be failing on Ruby 1.9.3 due to `Gem::InstallError: json_pure requires Ruby version ~> 2.0.`. This isn't something I touched, and I think would be failing in master too?

DLuCJ:
wrt json_pure:
https://github.com/puppetlabs/puppet/pull/5152

RickyCook:
Ah right okay. I assume I should ignore that failure then?

garethr:
@RickyCook thanks, much appreciated. I'd spotted the ability to pull via digest but not had time to add it to the module. Good job.

andrewsim:
may I ask when will this be released?

PR #583, opened 2016-08-31
Make selinux_enabled bool, instead of string

ofalk:
Adapt init and testsuite

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -448,9 +448,7 @@ class docker(
                 'log_driver must be one of none, json-file, syslog, journald, gelf, fluentd or splunk')
   }
 
-  if $selinux_enabled {
-    validate_re($selinux_enabled, '^(true|false)$', 'selinux_enabled must be true or false')
-  }
+  validate_bool($selinux_enabled)
 
   if $storage_driver {
     validate_re($storage_driver, '^(aufs|devicemapper|btrfs|overlay|overlay2|vfs|zfs)$',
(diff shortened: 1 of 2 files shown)


garethr:
In hindsight this is definitely correct, however as it stands this would be backwards incompatible (ie. it would break for anyone using it at the moment) which I'm loath to do for something this small.

However, I think if you change it to the following everything will be happy.

```puppet
validate_bool(str2bool($selinux_enabled))
```

This should fail for any old random string, but pass happily for a real boolean or for the strings "false" or "true". I've just tried that out briefly but adding a few additional tests would be useful to confirm that.

Thanks again, improvements like this to the interface are much appreciated. And apologies for taking a while to get back to the issue.

LongLiveCHIEF:
We could possibly change the default for this parameter `false`, but now that support for priveleges and such are much more stable, especially on RHEL systems, it's a poor choice for a default.

Setting it to default true would cause problems for compatibility on RHEL/CentOS majversion 6 OS's, since the latest docker packages available there are 1.7 (which don't have selinux support).

Overall, this one probably needs some additional logic in `manifests/params.pp` to keep from having to enter this param as a string.

LongLiveCHIEF:
aaaand I just saw @garethr 's comments.  I canceled the review my comments triggered. I landed on the diff page when searching the repo for `selinux_enabled` issues/MR's.

Is there a backwards compatible way to do something like the puppet 4 `Variant` type validator?:

```
class docker (
  Variant[Boolean,Undef],
)
```

ofalk:
Hi!
@garethr , could you implement this without me? I'm no longer working on that particular "project" here and therefore I'd have a hard time in testing it :-/
KR,
 Oliver

scotty-c:
Closing this PR due to inactivity, @ofalk if this is still needed please re-open the PR

Issue #505, opened 2016-04-21
Restart policy


vladgh:
`docker::run` validates the restart policies here: https://github.com/garethr/garethr-docker/blob/master/manifests/run.pp#L108

However, there is another policy that is not included: `unless-stopped`

Source: https://docs.docker.com/engine/reference/run/#restart-policies-restart

marji:
I have created PR #586 which adds the "unless-stopped" policy to the validation rule.

vladgh:
Thank you.

PR #586, opened 2016-09-04
#505 Added unless-stopped restart policy

marji:


marji:
I suspect the failed travis build above has nothing to do with my change, it failed with
"Gem::InstallError: json_pure requires Ruby version ~> 2.0.", while I just changed one string in puppet manifest.
I'll try to trigger another build by closing and reopening this pull requirest.

garethr:
@marji thanks, much appreciated.

Issue #588, opened 2016-09-12
email is deprecated in Docker > 1.13


dwerder:
In docker 1.13 the email flag for the registry login is deprecated, but the code in `registry.pp`  https://github.com/garethr/garethr-docker/blob/master/manifests/registry.pp#L46 requieres it.

jbehrends:
Just ran into this issue.. 
I'm running Docker v1.12.2 and attempted to call docker::registry passing in only a username and password.  As a result my puppet run failed with the following error:

Error: Cannot perform an interactive login from a non TTY device

If I include the email parameter and populate it with anything, puppet runs fine and configures docker with the provided username and password.

HadesArchitect:
Relates to https://github.com/garethr/garethr-docker/issues/657

fgabriel78:
On Docker 17.06, you get an error if you use the -email tag in your docker login command. So since Docker 17.06 it is impossible to login into a registry using Puppet :(
If you specify the email in docker::registry you get an error because the -email tag has been removed. If you don't specify the email in docker::registry you thet the "Cannot perform an interactive login from a non TTY device" error.

Houston, we have a problem!!

yaron-idan:
+1

joaquin386:
Encounter the same error.
Modify manually the registry.pp and remove all reference to the email.
After that the registry works perfectly I suppose an easy fix can be done for >17.06

But already done on the supported docker:
puppetlabs/puppetlabs-docker
So basicly you can close this case.

PR #577, opened 2016-08-17
Make email optional in docker login

dbason:
As per https://github.com/docker/docker/issues/6400 email is now not required for docker login

PR #589, opened 2016-09-12
allow registry login without email

dwerder:
Since docker 1.13 the email option on login is deprecated. Relates to issue #588


dwerder:
#577 addresses the same issue

garethr:
@dwerder @dbason thanks both, and apologies for taking an age to get to the issues. This looks like a good fix. Much appreciated.

PR #592, opened 2016-10-03
Add the memory cgroup to Grub on Debian systems

chriscowley:
Added support the option to enable memory cgroup support on Debian (enabled by default) along with an rspec test. Sorry, beaker is beyond me at the moment as I have (to my shame) only just started learning it. All I am doing is adding a line to `/etc/default/grub` and launching `update-grub`.

There is one problem: as it makes modifications to grub, it requires a reboot. Not sure how to work around this at present. Better minds than me please chime in.

As far as I know, this is only necessary on Debian itself, not Ubuntu.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -143,6 +143,7 @@ class docker::params {
       } else {
         $detach_service_in_init = true
       }
+      $manage_memory_cgroup = true
 
     }
     'RedHat' : {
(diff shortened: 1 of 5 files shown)


manifests/params.pp
```diff
@@ -143,6 +143,7 @@
       } else {
         $detach_service_in_init = true
       }
+      $manage_memory_cgroup = true
```
garethr: This will fail with strict variables on as the variable in question is referenced in the init but not set for other operating systems, You'll need to include it set to false for everything else I'm afraid.

PR #597, opened 2016-10-21
Specify additional environment files to add to the service-overrides.conf

emalihin:
It should be useful to include this commit to facilitate integration of Docker with other tools.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -307,6 +307,10 @@
 #   Specify a string for the docker group
 #   Default is OS and package specific
 #
+# [*daemon_environment_files*]
+#   Specify additional environment files to add to the
+#   service-overrides.conf
+#
 # [*repo_opt*]
 #   Specify a string to pass as repository options (RedHat only)
 #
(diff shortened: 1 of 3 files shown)


manifests/params.pp
```diff
@@ -28,6 +28,7 @@
   $selinux_enabled                   = undef
   $socket_group                      = undef
   $labels                            = []
+  $daemon_environment_files          = []
```
garethr: It doesn't look like this variable is used. The code is  little inconsistent at the moment with blank defaults. For the moment I'd just delete this line and leave the default in the init.pp/

garethr:
Thanks, nice idea and thanks for explaining the usecase. I left one picky comment to delete the line in params which isn't used, but happy to merge once that's done.

garethr:
@emalihin thanks. If you could squash the commits and rebase against master (as this now hits a merge conflict unfortunately) I'll get this merged and into the next release. Cheers

emalihin:
Thanks @garethr this should be ready now.

garethr:
@emalihin thanks, good job.

PR #599, opened 2016-10-28
Systemd restart policy

breser:
If on an systemd system and using the init style docker restart policies are not effective.  This is because as soon as the container stops for any reason systemd will see it exit and systemd executes the ExecStop commands which will cause docker to stop enforcing the restart policy.  Due to this behavior between the two it is not possible to use Docker restart policies with systemd.

Fortunately, systemd has it's own restart policies.  But current this is hardcoded in the template to on-failure.  So if you wanted the equivalent of the docker --restart=always in systemd you can't do it right now.

This pull request adds a new systemd_restart option that allows you to specify the systemd restart policy.  It defaults to on-failure so that the change has no impact on existing users.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -15,6 +15,11 @@
 #
 #    extra_parameters => ['--restart=always']
 #
+# However, if your system is using sytemd this restart policy will be
+# ineffective because the ExecStop commands will run which will cause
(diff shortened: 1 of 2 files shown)


garethr:
@breser thanks for the PR and for the explanation of why it's useful and for the comments in the code for anyone also wondering. Much appreciated.

PR #601, opened 2016-11-08
allow docker group to be overridden

adamcstephens:
This is really a workaround, but I'm not sure how I'd even tackle handling the params.pp logic without potentially impacting others.

I'm trying to use the RHEL7 stock docker, which uses group _dockerroot_. Unfortunately [this params block](https://github.com/garethr/garethr-docker/blob/v5.3.0/manifests/params.pp#L173-L177) negates the logic in [this params block](https://github.com/garethr/garethr-docker/blob/v5.3.0/manifests/params.pp#L201-L204) forcing RHEL7 to always use the `default_docker_group` of _docker_.

I don't really know how to fix the non-intuitive use of the same variable name in params and init, where logic exists for it in params but users can only override it later in the processing path through init.

I'm open to other suggestions on how to fix this problem, but for now this is broken for my use case.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -289,6 +289,10 @@
 #   Specify an array of users to add to the docker group
 #   Default is empty
 #
+# [*docker_group*]
+#   Specify a string for the docker group
+#   Default is OS and package specific
+#
 # [*repo_opt*]
 #   Specify a string to pass as repository options (RedHat only)
 #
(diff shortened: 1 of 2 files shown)


garethr:
Thanks, this looks good. The underlying code could do with some tidyup but this is the correct interface. Thanks

PR #606, opened 2016-12-01
Add SyslogIdentifier for systemd service

lazyfrosch:
So the service output can be properly identified in syslog


lazyfrosch:
There was an spec error in my environment, not related to my change as it seems...

garethr:
@lazyfrosch thanks for the addition, nice idea. I've fixed up the issue with the tests in Travis in master but I've run this locally. It needed one minor change to the tests and all looks good. Cheers

PR #609, opened 2016-12-06
Prevent exceeding systemd line length limit

saracen:
systemd has a line length limit of 2048 characters.
https://bugs.freedesktop.org/show_bug.cgi?id=85308

We recently stumbled across this problem when running a container that accepted a lot of configuration by environment variables.

This simply adds a new line break after every parameter.

### DIFF
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -50,7 +50,7 @@ module Puppet::Parser::Functions
 
     multi_flags = lambda { |values, format|
       filtered = [values].flatten.compact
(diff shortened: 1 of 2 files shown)


garethr:
It looks like this has a knock-on effect on the init script and causes one of the unit tests to fail?

```
when passing a links option
      should contain File[/etc/init.d/docker-sample] with content =~ / --link example:one --link example:two / (FAILED - 1)
```

This looks like it's just because the assertion assumes there is no lineending and continuation character so if you could update the test as well that would be appreciated. Thanks for the fix too.

saracen:
@garethr Done :)

garethr:
@saracen thanks, much appreciated.

Issue #619, opened 2017-01-13
Adding support for Docker service 'After' attribute in Unit file


eyal-lupu:
Hello

My environment (but this should be environment agnostic):
> O/S: CentOS 7
> Docker 1.12.2

The docker-engine RPM sets up a systemd unit file with After=network.target option. While this is usually enough for most cases it break on our environment as we require some other units to start.

The enhancement request in here is to add support to this Puppet module so one could configure alternate 'After' option in the unit override file (override files already exist as part of this module so we just need to add the option).

PR #534, opened 2016-06-17
Add 'docker::tag' to manage docker image tags.

gasparch:
This module allows to add and remove tags from local docker images.

PR #620, opened 2017-01-13
Adding support for Docker service 'After' attribute in Unit file

eyal-lupu:
This is a pull request for: https://github.com/garethr/garethr-docker/issues/619

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -438,6 +438,7 @@ class docker(
   $service_overrides_template        = $docker::params::service_overrides_template,
   $service_hasstatus                 = $docker::params::service_hasstatus,
   $service_hasrestart                = $docker::params::service_hasrestart,
+  $service_unit_after                = $docker::params::service_unit_after,
 ) inherits docker::params {
 
   validate_string($version)
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -344,6 +344,8 @@
 # [*storage_pool_autoextend_percent*]
 #   Extend the pool by specified percentage when threshold is hit.
 #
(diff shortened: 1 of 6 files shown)


gregohardy:
Hi @eyal-lupu. Thanks for your contribution! The PR is looking good, though we require a few things to move forward. 

1. There will need to be some spec tests to confirm your changes are behaving in the manner you expect. 

An example is https://github.com/garethr/garethr-docker/pull/534/files#diff-57c32a956fccf5af23aba2bce1d74d1d. 

2. The commits will need to be squashed before they can be merged.

If you have any questions just let us know.

eyal-lupu:
Hi @garethr 

thank you for your response. I an not familiar with Ruby at all but reverse engineering the existing tests it seems that adding a test similar to the below should work
```      
context 'with systemd service_unit_after' do
        let(:params) { { 'service_unit_after' => 'network-online.target' } }
        it { should contain_file(service_override_file).with_content(/After=network-online.target/) }
        end  
```

It does run but seems to be failing as the override files is missing:

>   2) docker on RedHat with systemd service_unit_after should contain File[/etc/systemd/system/docker.service.d/service-overrides.conf] with content =~ /After=network-online.target/
     Failure/Error: it { should contain_file(service_override_file).with_content(/After=network-online.target/) }
       expected that the catalogue would contain File[/etc/systemd/system/docker.service.d/service-overrides.conf]
     # ./spec/classes/docker_spec.rb:214:in `block (5 levels) in <top (required)>'


As I am really not familiar with that environment I am a bit blocked in here. If someone could enable the tests to create the file I will be more than happy to take that from there

PR #621, opened 2017-01-17
gpg check optional

alvarolmedo:


alvarolmedo:
I've checked in my dev env that this patch works fine.
What is the CI doing exactly?

garethr:
@alvarolmedo The issue with the tests is they are using strict variables, and `package_source_check` isn't always defined for all branches in `params.pp`. You'll need to define the variable in the other branches and set to undef. 

Some more details about that here for reference https://docs.puppet.com/puppet/latest/lang_variables.html#unassigned-variables-and-strict-mode

alvarolmedo:
What did you mean with the rest of branches?
I should do PRs in the rest of the branch of this repo?

gregohardy:
Hey @alvarolmedo. The branches are the case $::osfamily 'RedHat' specifically, and missing in the other osfamily elements. They must be defined in the base docker::params and set to undef. An example is say $storage_config_template.

scotty-c:
@alvarolmedo Is this PR still valid? As there has been no update on this since Feb.

alvarolmedo:
So sorry. I can't review the PR the last months.
Please give me a pair of days to reply.

PR #622, opened 2017-01-23
Set RestartSec to 5 for systemd service

apeeters:
This avoids hitting the StartLimitBurst and StartLimitInterval limits in case starting a docker container fails immediately because the dependent containers are not available yet.


gregohardy:
Thanks for the contribution!

PR #630, opened 2017-02-08
Stop hiding failed image builds

samuelson:
With the previous code, a failing image build is marked success by puppet. 

This change removes '1' as a valid exit code so that failed image builds will be correctly be marked as "Error". Since '0' is the default, there is no need for the "returns" parameter.

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -124,7 +124,7 @@ define docker::image(
       environment => 'HOME=/root',
       path        => ['/bin', '/usr/bin'],
       timeout     => 0,
-      returns     => ['0', '1'],
+      returns     => ['0', '2'],
       require     => File['/usr/local/bin/update_docker_image.sh'],
     }
   }
--- a/templates/update_docker_image.sh.erb
+++ b/templates/update_docker_image.sh.erb
@@ -2,7 +2,7 @@
 #
 # Pulls a docker image.
 # Returns 0 if there a change.
-# Returns 1 if there is no change.
+# Returns 2 if there is no change.
 # Returns 3 if something when wrong.
 #
 DOCKER_IMAGE=$1
@@ -16,7 +16,7 @@ if [[ -z $AFTER ]]; then
   exit 3
 elif [[ $BEFORE == $AFTER ]]; then
   echo "No updates to ${DOCKER_IMAGE} available. Currently on ${AFTER}."
-  exit 1
+  exit 2
 else
   echo "${DOCKER_IMAGE} updated. Changed from ${BEFORE} to ${AFTER}."
   exit 0


samuelson:
The bug was introduced with this commit: 
https://github.com/garethr/garethr-
docker/commit/e029316abdda2746fb76c738e9936e58c98d058c

@solarkennedy can you comment on why the exec treats 1 as a valid exit code? We've run into this issue many times in Puppet trainings and it's difficult to debug because failed image builds are hidden.

samuelson:
Can someone with write access rerun the build? It looks like a Travis issue.

samuelson:
Nm, it turns out that puppet-syntax 2.3.0 [now fails on deprecation notices](https://github.com/voxpupuli/puppet-syntax/blob/v2.3.0/lib/puppet-syntax/tasks/puppet-syntax.rb#L69)

solarkennedy:
@samuelson I think the comments in the script are clear: The purpose is to provide a way to make it so it could be used as an "onlyif" and have puppet "do nothing" and converge when there are no updates.

I do agree errors could be handled better. Maybe just a `set -e` ?

samuelson:
Oh sorry @solarkennedy  I hadn't looked the script. It seems like `docker build` doesn't follow those same exit code and our use case is not making it to that point in the if/else block. Would it make sense to wrap all of the `docker build` statements in script?

samuelson:
I'm happy to do the work to fix it if that approach makes sense.

solarkennedy:
In the case that the user is asking for a remote image, returning 1 (no updates to an image) or 0 (updates pulled). I think a good PR would be to look at the actual return code of docker pull and return 2 (something non 1,0) if the docker pull failed.

I agree that in the other cases (dir, file, tar) accepting a 1 is not appropriate. That was an oversight on my part with this PR.


The core problem I wanted to solve was convergence. If you don't wrap it like this, puppet will never converge (and always run docker build). I think there are 3 options:
1. (This PR) don't hide error codes for the "present" case and any build case
2. Wrap docker build in a script that returns 0/1 and do an onlyif trick on "present". (analogous to what I did for the docker pull use-case)
3. In the [case](https://github.com/garethr/garethr-docker/blob/671ad281c501b1d4706efd103b4aac00d5855078/manifests/image.pp#L97) switch also list out the acceptable return codes for each command. 
```
  if $docker_dir {
    $image_install = "${docker_command} build -t ${image_arg} ${docker_dir}"
    $returns = [0]
  } elsif $docker_file {
    $image_install = "${docker_command} build -t ${image_arg} - < ${docker_file}"
    $returns = [0]
  } elsif $docker_tar {
    $image_install = "${docker_command} load -i ${docker_tar}"
    $returns = [0]
  } else {
    $image_install = "/usr/local/bin/update_docker_image.sh ${image_arg}"
    $returns = [0,1]
  }
```

None of these options are super great, mostly because we are mixing 4 install methods (builds, load, pull) and 3 desired states (absent, latest, present) and each has its own nuances regarding onlyif's and return codes. This is going to be hard to get actually correct *and* allow puppet to converge. 

I think option 1 (this pr) is actually ok. In the remote image case: "latest" will still converge as before, "present" will only get exec'd once anyway (due to the unless).

Shipit.

samuelson:
The other option that occurred to me was changing the script to return some other exit codes. Maybe match the values of running puppet with the  '--detailed-exitcodes' flag?
0 = no changes
1 = fail
2 = Success with changes
etc.

solarkennedy:
Sounds good.

They key part there to making this work *and* giving puppet the ability to converge is enabling is providing the right "onlyif" script that returns 0 when there are no changes to. We are basically abusing the fact that onlyif suppresses the Exec to allow puppet to converge.

samuelson:
Ok, cool, I'll take a crack at it on Tuesday when I'm back at work. Thanks for the feedback.

samuelson:
Ok, I just changed it to 2 and left the other fail exit code at 3. That way it doesn't conflict with the exit code 1 from `docker build` if that ever gets used in the future.

samuelson:
@gregohardy @solarkennedy how does that look?

gregohardy:
@solarkennedy I've kicked travis CI to get it green after my push. I'm good with the changes. If you +1 it I'll merge.

solarkennedy:
lgtm

gregohardy:
Cheers for the contribution @samuelson

samuelson:
Thanks!

PR #638, opened 2017-02-28
update to https

elliot-labs:
Upgraded the key source to https communication.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -133,10 +133,10 @@ class docker::params {
       $storage_setup_file = undef
 
       $package_cs_source_location = 'http://packages.docker.com/1.9/apt/repo'
-      $package_cs_key_source = 'http://packages.docker.com/1.9/apt/gpg'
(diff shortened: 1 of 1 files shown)


scotty-c:
LGTM

garethr:
@scotty-c I'd probably revert that without a lot more testing, there is a history of bugs hidden behind making that change. eg. https://github.com/garethr/garethr-docker/issues/369#issuecomment-180288521. The short part is with this change the module has an unspecified dependency.

I'd also like more context for the why. In the majority of cases it's not better for most values of better I believe, eg. https://askubuntu.com/questions/146108/how-to-use-https-with-apt-get and https://askubuntu.com/questions/352952/are-repository-lists-secure-is-there-an-https-version

scotty-c:
@garethr I have done extensive testing with this change. Docker itself suggests using `https` for its repos. An example of this are https://docs.docker.com/engine/installation/linux/debian/#install-using-the-repository and is also heavily used in Docker install scripts at https://get.docker.com/ As the module is in your namespace, I am more than happy to revert the commit if that is what you would prefer.

PR #639, opened 2017-02-28
depreciation update

elliot-huffman:
Based upon depreciation messages, syntax was updated to reflect new commands.

tested and working on my system.


elliot-labs:
I have updated the PR to pass the build check.

PR #641, opened 2017-03-02
Limit Image -> Run dependencies to those used

### LINKED ISSUE #640: Docker::Run will depend on all non related Docker::Images within the same node
Using puppet to bring up a number of containerized services on a single node has resulted in a few cyclical dependency issues. In reviewing this (and resolving our own mistakes) noticed a rather odd dependency situation where every Docker::Run[...] will depend on all Docker::Image[...] resources declared irrespective of relevancy.

I've managed to trim this down to a minimal site manifest for a single node that from which a relationship.dot graph can be generated showing the issue:
```
node 'srv-2.example.net' {
  class { '::docker':
    use_upstream_package_source => false,
    package_name                => 'docker.io',
  }

  ::docker::image { 'openfrontier/gerrit:latest':
    ensure  => 'latest',
  }

  ::docker::run { 'gerrit':
    ensure => 'present',
    image  => 'openfrontier/gerrit:latest',
    net    => undef,
    ports  => [
      '8080:8080',
      '29418:29418',
    ],
    subscribe => Exec["echo 'Update of openfrontier/gerrit:latest complete'"],
    privileged => false,
  }

  ::docker::image { 'nginx/stable-alpine:latest':
    ensure  => 'latest',
  }

  ::docker::run { 'nginx':
    ensure => 'present',
    image  => 'nginx/stable-alpine:latest',
    net    => undef,
    ports  => [
      '80:80',
    ],
    subscribe => Exec["echo 'Update of nginx/stable-alpine:latest complete'"],
    privileged => false,
  }
}
```

Then generating a dependency graph with:
```
puppet agent \
    --server <master> \
    --verbose \
    --ignorecache \
    --no-usecacheonfailure \
    --no-splay \
    --noop \
    --no-daemonize \
    --onetime \
    --noop \
    --graph
```

Then use `xdot` to view the graph

Since the .dot files are plain text, it's possible to look directly at the contents of relationship.dot and see an unexpected set of dependencies being added:
```
    "Class[Docker]" -> "Docker::Image[openfrontier/gerrit:latest]" [
        fontsize = 8
    ]

    "Class[Docker]" -> "Docker::Run[gerrit]" [
        fontsize = 8
    ]

    "Docker::Image[openfrontier/gerrit:latest]" -> "Docker::Run[gerrit]" [
        fontsize = 8
    ]

    "Docker::Image[openfrontier/gerrit:latest]" -> "Docker::Run[gerrit]" [
        fontsize = 8
    ]

    "Exec[echo 'Update of openfrontier/gerrit:latest complete']" -> "Docker::Run[gerrit]" [
        fontsize = 8
    ]

    "Docker::Image[nginx/stable-alpine:latest]" -> "Docker::Run[gerrit]" [
        fontsize = 8
    ]

    "Docker::Image[nginx/stable-alpine:latest]" -> "Docker::Run[gerrit]" [
        fontsize = 8
    ]

    "Class[Docker]" -> "Docker::Image[nginx/stable-alpine:latest]" [
        fontsize = 8
    ]

    "Class[Docker]" -> "Docker::Run[nginx]" [
        fontsize = 8
    ]

    "Docker::Image[openfrontier/gerrit:latest]" -> "Docker::Run[nginx]" [
        fontsize = 8
    ]

    "Docker::Image[openfrontier/gerrit:latest]" -> "Docker::Run[nginx]" [
        fontsize = 8
    ]

    "Docker::Image[nginx/stable-alpine:latest]" -> "Docker::Run[nginx]" [
        fontsize = 8
    ]

    "Docker::Image[nginx/stable-alpine:latest]" -> "Docker::Run[nginx]" [
        fontsize = 8
    ]

    "Exec[echo 'Update of nginx/stable-alpine:latest complete']" -> "Docker::Run[nginx]" [
        fontsize = 8
    ]
```
Notice that while the expected dependencies of the run depending on a corresponding image:
```
"Docker::Image[openfrontier/gerrit:latest]" -> "Docker::Run[gerrit]"
and
"Docker::Image[nginx/stable-alpine:latest]" -> "Docker::Run[nginx]"
```

There is the additional unexpected dependencies:
```
"Docker::Image[nginx/stable-alpine:latest]" -> "Docker::Run[gerrit]"
and
"Docker::Image[openfrontier/gerrit:latest]" -> "Docker::Run[nginx]"
```

In a larger deploy this creates problems when trying to group services into their image + run actions in a separate manifest and then if you add any dependencies between those services such as having the nginx manifest depend on gerrit manifest you'll trigger a dependency cycle because the gerrit docker::run is already depending on the nginx docker::image.

Recognize this is a slightly contrived example, but it's not unusual for people to wrap all the pieces for one service in a dedicated manifest and then create dependencies between these top level services. The idea that somehow there is a cyclical dependency then between them catches most by surprise as there does not appear to be a good reason for the additional dependencies between the run and image.

electrofelix:
Use searching in the resource collection to limit adding dependencies on
Docker::Image for only those Docker::Run resources defined to use the
specific image.

This reduces the likelihood of accidental cyclical dependencies where
consumers defined a manifest per service containing an image and run,
and then try to add a dependency from one service class onto another.

Fixes #640

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -129,4 +129,5 @@ define docker::image(
     }
   }
 
+  Docker::Image <| title == $title |> -> Docker::Run <| image == $image_arg |>
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -527,7 +527,7 @@ class docker(
   contain 'docker::config'
   contain 'docker::service'
 
-  Class['docker'] -> Docker::Registry <||> -> Docker::Image <||> -> Docker::Run <||>
-  Class['docker'] -> Docker::Image <||> -> Docker::Run <||>
+  Class['docker'] -> Docker::Registry <||> -> Docker::Image <||>
+  Class['docker'] -> Docker::Image <||>
   Class['docker'] -> Docker::Run <||>
 }
--- a/spec/acceptance/docker_full_spec.rb
+++ b/spec/acceptance/docker_full_spec.rb
@@ -541,6 +541,32 @@ describe 'the Puppet Docker module' do
 
         shell('docker inspect container-3-6', :acceptable_exit_codes => [1])
       end
+
+      it 'should allow dependency for ordering of independent run and image' do
+        pp=<<-EOS
+          class { 'docker':}
+
+          docker::image { 'ubuntu': }
+
+          docker::run { 'container_3_7_1':
+            image   => 'ubuntu',
+            command => 'init',
+          }
+
+          docker::image { 'busybox':
+            require => Docker::Run['container_3_7_1'],
+          }
+
+          docker::run { 'container_3_7_2':
+            image   => 'busybox',
+            command => 'init',
+          }
+        EOS
+
+        apply_manifest(pp, :catch_failures => true)
+        apply_manifest(pp, :catch_changes => true) unless fact('selinux') == 'true'
+
+      end
     end
 
     describe "docker::exec" do


electrofelix:
I'll have to come back to adding a test. It looks like the only way to do this is with an acceptance check, however really if I create 2 custom classes with a different image and run entry to represent two services and define a dependency between the two classes this change should prevent a dependency cycle from being created and allow for image1 -> run1 -> image2 -> run2 to occur without an issue.

If you have any suggestions on whether it's possible to add that to the unit tests it would be very useful since this really only needs to check puppet parsing doesn't see a dependency cycle rather than needing to actually apply the puppet code.

electrofelix:
Hope this simple acceptance test which fails before my previous change due to a dependency cycle and passes with it is sufficient for this PR to be approved.

electrofelix:
Force pushed to trigger a recheck

electrofelix:
@gregohardy who needs to review this? any way I can help move it forward?

electrofelix:
@garethr @gregohardy is this acceptable to merge yet

electrofelix:
Appears this repository has been forked and work is continuing over at puppetlabs/puppetlabs-docker

Resubmitted as puppetlabs/puppetlabs-docker#199

PR #644, opened 2017-03-09
adding swarm mode

scotty-c:
This PR adds Swarm mode. It contains 2 custom parsers. One to create the cluster and the other to join the cluster. In the class docker::swarm we have the configuration for the define type, there is validations on all inputs. By default, the class is not included if you just call the parent class ```include docker``` This was modelled off the existing ```docker::compose``` class

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,6 +466,40 @@ docker_compose { '/tmp/docker-compose.yml':
 It is also possible to give options to the ```docker-compose up``` command
 such as ```--remove-orphans``` using the ```up_args``` option.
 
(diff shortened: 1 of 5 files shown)


drignoto:
@scotty-c Great! I'm still develop some changes in your branch for export tokens via facter and use the compose v3 version  via `docker stack`. In a couple of days you have the pull request ;)

scotty-c:
@drignoto Thanks for that !!! I had actually got that work in my backlog of things to do. So your help is greatly appreciated.

PR #652, opened 2017-03-15
Add support for onlyif in a docker exec

electrofelix:
Allow adding a condition to only run the exec in specific cases.
Additionally provide a simple 'running' option that allows checking if
the container is running before attempting an exec in cases where it may
be used to perform config testing before allowing a restart of a
container, but should be skipped if the container is not running.


electrofelix:
anyone interested in reviewing/merging this?

electrofelix:

PR #653, opened 2017-03-16
Support refreshonly flag

electrofelix:
Allow for the refreshonly flag to be passed through to the underlying
exec resource.

This allows for Docker::Exec resources to be defined to only run when
notified, which allows for construction of manifests that can check
whether updated configuration made available to running containers
(provided they are using directory mount binds) are valid before
triggering a service restart.

Use case for this is to test generated nginx configuration for the
running nginx container using a Docker::Exec call that is notified by
the file resource and in turn will notify the service resource to
restart.

### DIFF
--- a/manifests/exec.pp
+++ b/manifests/exec.pp
@@ -8,6 +8,7 @@ define docker::exec(
   $tty = false,
   $container = undef,
   $command = undef,
+  $refreshonly = false,
   $unless = undef,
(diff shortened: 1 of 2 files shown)


electrofelix:
Anyone interested in reviewing/merging this?

electrofelix:

PR #658, opened 2017-03-21
Add support for fetching docker-compose from alternate URLs

jamesmcdonald:
I have added parameters to allow overriding of either the "base" URL or the whole URL when installing docker-compose.

Using baseurl allows you to install from a local mirror of whatever version/architectures you need with the same directory layout as the repository on GitHub.

Using rawurl you can override the entire URL and point to a specific file, dealing with the version and architecture yourself.

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -20,17 +20,37 @@
 # [*proxy*]
 #   Proxy to use for downloading Docker Compose.
 #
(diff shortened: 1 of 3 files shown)


jamesmcdonald:
Ahh, it looks like #456 already has a solution for this. I'll leave it open and you can decide what you want to do with it.

PR #663, opened 2017-04-05
Add parameter docker::service_autorestart to avoid docker daemon restarts

mrodm:
Currently, if there is any change in any of configuration files of docker (e.g. /etc/default/docker), this will notify to the service resource, and thus restarting the docker daemon out of our control (if it would be needed). 

This PR adds a new boolean parameter `service_autorestart` allowing us to decide whether or not service should be restarted when configuration files are modified.


mrodm:
Travis fails when downloading dependencies of ruby gems, it should not be related to this PR

```
Fetching gem metadata from https://rubygems.org/...Retrying dependency api due to error (2/3): Bundler::HTTPError Net::HTTPClientError: Too many gems! (use --full-index instead)

Retrying dependency api due to error (3/3): Bundler::HTTPError Net::HTTPClientError: Too many gems! (use --full-index instead)
```

mrodm:
I've tried to run the tests in master branch, and syntax test also fails in that branch

PR #666, opened 2017-04-09
simple compose v3, v3.1, v3.2 support

joaquin386:
Simple solution to make docker compose v3 yaml file work with garethr-docker
More things are needed like modifying the dokcer run with the new v3 attributes, next job.

### DIFF
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -16,7 +16,7 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
       "label=com.docker.compose.project=#{project}"
     ]).split("\n")
     counts = case compose_file["version"]
-    when /^2(\.0)?$/, /^3(\.[0-1])?$/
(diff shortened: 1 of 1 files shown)


scotty-c:
Hey @joaquin386 
Thanks for this PR, would you be able to make a small change so it also supports docker compose version `3.1` as listed here https://docs.docker.com/compose/compose-file/compose-versioning/
Once the PR has that functionality it should be good to merge

joaquin386:
When /^3(\.[0-1])*$/
Just putting down the change needed and for future change in case a 3.X is out.
This also goes with the version of docker-compose. By default on this module it is 1.7 and 3.1 should go with 1.12 (but this should be done more carefull since certain tags got deprecated from 2 to 3)

dbason:
:+1

joaquin386:
Added the new Line for support for docker-compose 3 and 3.1

when /^2(\.0)?$/, /^3(\.[0-2])*$/

scotty-c:
Thanks for the PR !!!

PR #668, opened 2017-04-11
WIP: Add facts from docker version and docker info.

mterzo:
Add structured facts to support docker version and docker info.


mterzo:
I wanted to start a conversation on this, by no means is it complete.  Look for some feed back.

PR #679, opened 2017-05-05
COUD-154 Adding Docker services

scotty-c:
Adding functionality to use Docker services https://docs.docker.com/engine/reference/commandline/service_create/ within a swarm

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,6 +499,53 @@ docker::swarm {'cluster_worker':
 ensure => absent
 }
 ```
+### Docker services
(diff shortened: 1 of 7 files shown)


scotty-c:
I have used a Puppet 4 function in the PR. So the tests are failing for Puppet 3, I will make a change to make sure Puppet 3 are still supported

scotty-c:
I have removed the Puppet 4 function, to allow compatibility for both Puppet 3 and 4. This is now ready to merge

MWilsonPuppet:
@scotty-c This looks good to me Scott.

Issue #637, opened 2017-02-22
world readable initscripts are a security vulnerability


mujiburger:
This is similar to the problem addressed by 8fa788be.  The initscripts created by `docker::run()` are world readable/executable by default which exposes potentially sensitive env variables to any authenticated system user.

Setting their `owner` to `"root:${docker::params::docker_group}"` and `mode` to `'0750'` or `'0640'` depending on the type of file should resolve this problem.  I am happy to submit a PR with these changes if you think that is an acceptable way to mitigate the problem.

scotty-c:
@mujiburger If you could submit a PR with your suggested changes it would be greatly appreciated.

PR #680, opened 2017-05-08
Tighten up init script permissions

mujiburger:
Re: #637

Obviously, implementing docker secrets in this module is the correct solution to the problem. But this PR helps mitigate a potential credential leak in the current code.


scotty-c:
@mujiburger I will test this one out tonight

scotty-c:
Thanks for your contribution to the project @mujiburger

PR #681, opened 2017-05-12
Make docker::run use $service_name instead of "docker"

benningm:
When using the module with 'docker-latest' packages from RedHat the systemd service of docker is also 'docker-latest'.
There is already a $service_name variable for this, but 'docker' is still hardcoded in some strings.

This replaces hardcoded 'docker' strings with $service_name.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -117,7 +117,7 @@ define docker::run(
   }else {
     $docker_command = $docker::params::docker_command
   }
-  $service_name = $docker::params::service_name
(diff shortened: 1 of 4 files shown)


gdlx:
Your PR is missing service overrides template:
https://github.com/garethr/garethr-docker/blob/master/templates/etc/systemd/system/docker.service.d/service-overrides-rhel.conf.erb#L2

benningm:
Just added the service overrides and resynced with upstream

Issue #607, opened 2016-12-04
Apt::Source and Apt::Key deprecation warnings in garethr-docker v5.2.0


CodeGnome:
The garethr-docker module v5.2.0 is installed by puppetlabs-docker_platform v2.1.0. It generates a number of deprecation warnings on Puppet 3.8.5 when running `class { 'docker': }` or `docker::image`, including:

> Warning: Config file /etc/puppet/hiera.yaml not found, using Hiera defaults
> Warning: Scope(Apt::Source[docker]): $include_src is deprecated and will be removed in the next major release, please use $include => { 'src' => false } instead
> Warning: Scope(Apt::Source[docker]): $required_packages is deprecated and will be removed in the next major release, please use package resources instead.
> Warning: Scope(Apt::Source[docker]): $key_source is deprecated and will be removed in the next major release, please use $key => { 'source' => http://apt.dockerproject.org/gpg } instead.
> Warning: Scope(Apt::Key[Add key: 58118E89F3A912897C070ADBF76221572C52609D from Apt::Source docker]): $key_source is deprecated and will be removed in the next major release. Please use $source instead.

On the surface, it looks like the errors are being thrown by the APT module, but the source of the problem is actually in the way that the Docker module is invoking Apt::Source and Apt::Key.

covataamos:
Still there for version 5.3.0

dpash:
As of puppetlabs-apt 4.0.0, this now throws an error.

bittner:
@garethr There are now at least 6 PRs that try to fix this very issue:

1. #461
1. #528
1. #595
1. #634
1. #675
1. #693

None has been merge until today. Why?

eliaoggian:
+1 @bittner 

Please merge the fix.

BlauerPulli:
my ubuntu 16.04 VM with puppet 3.8.5 and the garethr-docker v5.3.0 generates no warnings but an error:  
_Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Puppet::Parser::AST::Resource failed with error ArgumentError: Invalid resource type apt::source at /etc/puppet/modules/docker/manifests/repos.pp:34 on node puppetagentone.local_

I've already done @costela s changes- still the same error..

BlauerPulli:
info: now i use puppet 4 and the official "docker_platform" puppet module(still on ubuntu 16.04) and it works.

bittner:
[...]

PR #675, opened 2017-04-28
Replace deprecated apt::source parameters

dpash:
A number of apt:source parameters were deprecated, and have now been
removed. This commit migrates to replacements

* key_source is replaced with passing a hash to the key parameter.
* include_src is replaced with passing a hash to the include parameter.
* required_packages has been replaced with explicitly installing them,
and adding a dependency between the package and the apt::source.

I'm not sure if the explicit package installation might cause an issue
with other modules doing the same thing.

This will resolve #607

PR #595, opened 2016-10-12
Fix deprecated apt::source parameters

adepretis:
`puppetlabs/apt` 2.0.0 has been release well over a year ago and deprecated some parameters for `apt::source` (see [Changelog](https://forge.puppet.com/puppetlabs/apt/changelog#aptsource)).

In `puppetlabs/apt` version 2.0.0 these parameters (`key_source`, `include_src`, ...) have been completely removed, in version >= 2.1.0 using these results in a deprecation warning.

Isn't it time to finally implement these changes and require `puppetlabs/apt` >= 2.0.0? ;-)

PR #693, opened 2017-06-01
fixes a bug introduced by changes upstream in the apt module.

bobweston:
This Pull Request addresses bug introduced by updated apt module.

It appears that ruby 1.9.3 needs to be removed from travis along with STRICT_VARIABLES is no longer supported in APT Module v2.1

That seems to be the cause of the failed tests

See #607 for more detail on what this addresses.

### DIFF
--- a/lib/facter/docker.rb
+++ b/lib/facter/docker.rb
@@ -1,0 +1,30 @@
+require 'facter'
+require 'json'
+
+def interfaces
+  Facter.value(:interfaces).split(',')
+end
+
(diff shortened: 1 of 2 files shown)


Cosaquee:
#634 #595 #461 #528 all address the same issue. @garethr can you maybe close the duplicates after choosing one PR that fits you and tell us what should be done to merge fix like this?

bittner:
Same for #675. @garethr is there a chance to act on this problem?

PR #695, opened 2017-06-09
Use line breaks and continuation for run flags

### LINKED ISSUE #660: systemd files exceeding MAX_LINE
─┬ garethr-docker (v5.3.0)
Puppet 4.8.1

Due to the concatenation of the environment variables the max line length supported by systemd is exceeded when multiple environment variables are used. This causes strange errors in executing the puppet file that give no hint to the root cause.
To prevent that every environment variable should be in a single line.

electrofelix:
Switch from joining the docker_run_flags using spaces to new lines and
continuations to avoid hitting the glibc MAX_LINE limit used by systemd.

As docker_run_flags is used to provide the args to init scripts and
systemd, switching it to use line continuation and newlines is
sufficient to ensure that systemd/initscritps do not hit any limit
imposed by glibc on a readline and at the same time know to continue
processing onto the next line for the rest of the command to be run.

Fixes #660

### DIFF
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -76,6 +76,8 @@ module Puppet::Parser::Functions
       flags << param
     end
 
-    flags.flatten.join(" ")
+    # Some software (systemd) will truncate very long lines using glibc's max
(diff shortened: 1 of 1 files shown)


electrofelix:
@garethr Does this look ok to merge?

electrofelix:

PR #697, opened 2017-06-13
When not given any tag, remove all tags

gdubicki:


gdubicki:
Frankly speaking I think that I should rather make it configurable, opt-in behaviour as the way it currently works mirrors docker cli `docker rmi`, which when not given any tag tries to delete `latest`.

Issue #701, opened 2017-06-30
Docker Community Edition `docker daemon` is not supported on Linux.


ocastle:
Encountered the following error on Ubuntu 16.04 running docker version 17.06.0-ce.

``docker daemon` is not supported on Linux. Please run `dockerd` directly`

The ExecStart= line of the 'service-overrides-debian.conf.erb' template needs to use 'dockerd' as the command instead of 'docker daemon' when using the Community Edition. Editing the value and reloading the systemd daemon confirmed. Puppet run will revert the change, but daemon will continue running until  restarted and require the same change to restore service.

grantholly:
@ocastle would you be willing to test out some changes that I have in a fork?  I've been having the same problem with CentOS and docker 17.06.0-ce, but I've hacked together a workaround.  Its in my fork https://github.com/grantholly/garethr-docker.

vladgh:
From my tests, the latest docker (`docker-ce_17.06.0~ce-0~ubuntu`) on Xenial does not use the `/etc/systemd/system/docker.service.d/service-overrides.conf` file. The workaround for me was to delete that file altogether, rather than modifying `/usr/bin/docker daemon` to `/usr/bin/dockerd`.

I'm not sure, but I think `/lib/systemd/system/docker.service` is used instead.

ocastle:
@grantholly yes i'd be willing to take a look at the changes and integrating into my own fork for testing.

Morsicus:
Hi,

Any news on this issue ? :)

ocastle:
The changes from @grantholly were integrated into the working fork used for testing and were verified to solve the issue faced on Ubuntu:16.04

LongLiveCHIEF:
> The workaround for me was to delete that file altogether, rather than modifying /usr/bin/docker daemon to /usr/bin/dockerd.

This is related to #703 & #698. You actually _shouldn't_ have to manually perform any workarounds, and should instead be able to pass in `dockerd` as the `docker_command`, but it's not writing out to the correct file.

juliantaylor:
In addition to changing `docker_command` to `dockerd`, you also need to override `daemon_subcommand` to an empty string to get a correct service file override. See https://github.com/garethr/garethr-docker/blob/master/templates/etc/systemd/system/docker.service.d/service-overrides-debian.conf.erb

LongLiveCHIEF:
[...]

PR #704, opened 2017-07-10
support dockerd and docker daemon in unit files

grantholly:
Using `docker daemon` has been deprecated since docker v1.13.  This means that if you are using a recent version of docker, the unit files for upstart and systemd call the deprecated binary that is not installed.  However, for people running older versions of docker, using `docker daemon` is correct.  In this pull request I am adding support for people using `dockerd`.  Here is a related issue https://github.com/garethr/garethr-docker/issues/701.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,10 +1,14 @@
 fixtures:
-  forge_modules:
-    stdlib: 'puppetlabs-stdlib'
-    epel: 'stahnma-epel'
+#  forge_modules:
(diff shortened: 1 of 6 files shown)


Morsicus:
Works well on docker >= v1.12.1

vide:
I'm no mantainer but, couldn't this be more easily done just assigning proper default values in `params.pp` to `docker::service::docker_command` and `docker::service::daemon_subcommand` ?

PR #705, opened 2017-07-12
Add support for the proper --cpuset-cpus flag

### LINKED ISSUE #448: Flag '--cpuset' doesn't exist
Flag '--cpuset' doesn't exist with Docker 1.10.
Output when trying to start: "
Starting docker-elasticsearch:	flag provided but not defined: --cpuset"

From init-script: "
    printf "Starting $prog:\t"
                $docker rm  elasticsearch >/dev/null 2>&1
            $docker run \
    --net bridge -m 0b --cpuset=0-7
..<snip>"

Hiera-conf: "
docker::run_instance::instance:
  elasticsearch:
    image: 'elasticsearch:2.2'
    cpuset: '0-7'
...<snip>"

I think it should be '--cpuset-cpus' according to the docs.

rogierslag:
The old one was removed in Docker 1.10.0 so systems would no longer properly boot after an upgrade. This restores the behaviour with the same API for the puppet type

Closes https://github.com/garethr/garethr-docker/issues/448

### DIFF
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -29,7 +29,7 @@ module Puppet::Parser::Functions
     cpusets = [opts['cpuset']].flatten.compact
     unless cpusets.empty?
       value = cpusets.join(',')
-      flags << "--cpuset=#{value}"
+      flags << "--cpuset-cpus=#{value}"
     end
 
     if opts['disable_network']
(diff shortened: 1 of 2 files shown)


rogierslag:
Note the PR is much simpler than https://github.com/garethr/garethr-docker/pull/460 on purpose. This way the external API is consistent and we abstract over the internal API of Docker. Adding `--cpuset-mems` might become confusing though (the docker command is in my opinion), but since that is not (yet) in here, I decided not to go for that.

I couldnt quickly figure out how to make this work for Docker < 1.10 though, so it would mean breaking those clients. To support this, we might want to create a facter fact to allow more modules to depend on the current docker version

PR #708, opened 2017-07-28
Set DOCKERBINARY in sysconfig to docker_command

rochaporto:
This will allow compatibility with the centos packages, where:
/usr/bin/docker: a script sourcing /etc/sysconfig/docker

and docker-client and docker-client-latest packages install
/usr/bin/docker-current and /usr/bin/docker-latest, respectively.

Setting DOCKERBINARY in sysconfig defines which of the pkgs should be
used.

### DIFF
--- a/templates/etc/conf.d/docker.erb
+++ b/templates/etc/conf.d/docker.erb
@@ -2,6 +2,7 @@
 # may be overwritten
 
 DOCKER="/usr/bin/<%= @docker_command %>"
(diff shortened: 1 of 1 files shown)


rochaporto:
An alternative for us is to set this directly in sysconfig/docker, without using the module.

Is this PR something you would be interested in taking, or should we do the alternative instead?

rochaporto:
Any feedback on this?

PR #713, opened 2017-08-30
Updates apt::source usage

amateo:
The use of key_source in apt::source is deprecated since puppetlabs/apt
version 2.1.1 (at least) and removed in version 4.0.0.


noqqe:
Please merge, unless this is in the codebase, there is no way installing the module with `librarian-puppet`. 

`apt` dependency is way too old.

jarro2783:
+1

PR #733, opened 2017-10-20
Minor change - Uncommented TMPDIR for rhel 7 machines

Colza5000:
This is causing issues as we (and i assume a majority of others) set `noexec` on `/tmp` and it's causing docker to produce this error.

```docker-compose: error while loading shared libraries: libz.so.1: failed to map segment from shared object: Operation not permitted```


soundstep:
I need this to be fixed as well.
What's the state for that PR? Are you considering it? Is it been looked at?
Status update please ;)
Cheers.

PR #734, opened 2017-10-25
fixed module to support new syntax of puppetlabs-apt

devfaz:
Hi,

just tried to fix the module to support the new puppetlabs-apt syntax.

 Fabian


devfaz:
Well, nobody cares.

PR #745, opened 2018-01-18
Add refreshonly feature to docker::exec

ofalk:
Very simple change to add refreshonly feature so docker::exec is only run if it's notified by something else.


ofalk:
@garethr - anything wrong with this request?

ofalk:
Ping @garethr !?

PR #756, opened 2018-07-02
Allow overriding service_provider on docker::run

### LINKED ISSUE #755: $docker::service_provider is not respected by docker::run
`docker::service_provider` can be manually set to override the choice of service provider for the docker daemon itself, but containers created with `docker::run` do not respect this and use hardcoded defaults for particular operating systems. `params.pp` already does the work of figuring out whether the node is running systemd (which can be overridden when necessary).

I propose to add a single check to see if `docker::service_provider` is systemd, and if so set the necessary variables once (rather than again, duplicating those for each OS which does support systemd), and fall back to the OS-specific blocks for non-systemd OSes.

Corresponding pull request incoming.

optiz0r:
Its possible to override the `service_provider` when managing the docker
daemon in case the defaults in params.pp aren't correct, but not when
defining a container using `docker::run`. This change adds a new
`service_provider` parameter to the run defined type, so that containers
can also be managed in these situations.

Fixes garethr/garethr-docker#755

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -81,6 +81,7 @@ define docker::run(
   $dns = [],
   $dns_search = [],
   $lxc_conf = [],
+  $service_provider = undef,
   $service_prefix = 'docker-',
   $restart_service = true,
   $restart_service_on_docker_refresh = true,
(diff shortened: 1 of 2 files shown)


optiz0r:
Implemented in new upstream by https://github.com/puppetlabs/puppetlabs-docker/commit/a77179b52f2818e9e52059233c4f80725107892d

